PR #2, opened 2017-02-20
Refactor run/semantics to let the user choose which monad to run the

stevana:
semantics in.

### DIFF
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -38,9 +38,6 @@ data MemStepF ref
 
 type MemStep = MemStepF Ref
 
-newtype Mem = Mem { unMem :: [MemStep] }
-  deriving (Show, Eq, Monoid, Read)
-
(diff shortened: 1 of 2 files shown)


test/Spec.hs
```diff
@@ -124,62 +121,59 @@ instance Show Response where
...
   where
-  sem :: OurMonad m => Mem -> m ()
-  sem = foldM (\ih ms -> semStep' ms >> return ih) () . unMem
+  semStep'
+    :: (MonadIO m, MonadState Int m)
+    => MemStep -> StateT [IORef Int] m Response
+  semStep' = liftSem (semStep None) isRef
+
+  semSteps :: MonadIO m => [MemStep] -> m [IORef Int]
+  semSteps = flip evalStateT 0 . flip execStateT [] . go
```
Danten: Does this actually type-check.. how?

PR #3, opened 2017-02-20
There isn't any need to scopecheck in parallel property helper.

stevana:


test/Spec.hs
```diff
@@ -251,6 +251,13 @@ prop_parallel prb = parallelProperty
 
 ------------------------------------------------------------------------
 
+scopeCheck :: RefKit cmd ref => [cmd ref] -> Bool
+scopeCheck = go 0
+  where
+  go _ []       = True
+  go s (c : cs) = all (\r -> r < toEnum s) (usesRefs c) &&
+    go (if returnsRef c then s + 1 else s) cs
+
```
Danten: We could move `prop_genScope`, `prop_shrinkForkScope`, `prop_shrinkForkSubseq` to be generic test-of test functions... Actaully I see now that some of them uses `cheat` which is a bit annoying...

stevana: Yeah, that would be nice -- added it to the to do list.

PR #6, opened 2017-02-21
Remove scope/index from response; we can calculate it when pretty

stevana:
printing.

### DIFF
--- a/src/Test/QuickCheck/StateMachineModel.hs
+++ b/src/Test/QuickCheck/StateMachineModel.hs
@@ -1,8 +1,8 @@
-{-# LANGUAGE DeriveFunctor          #-}
-{-# LANGUAGE FlexibleContexts       #-}
-{-# LANGUAGE MultiParamTypeClasses  #-}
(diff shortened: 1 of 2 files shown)


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -299,20 +299,29 @@ linearise StateMachineModel {..} = map fromRose . filter (postConditionsHold ini
...
   (l, r)  = partition (\e -> getProcessId e == 1) h'
 
-  mkOps :: [HistoryEvent cmd resp] -> [Operation cmd resp]
-  mkOps [] = []
-  mkOps (InvocationEvent cmd _ : ResponseEvent resp pid : es)
-    = Operation cmd resp pid : mkOps es
-  mkOps _ = error "mkOps: Impossible."
+  (p', n) = mkOps 0 p
+
+  mkOps :: Int -> [HistoryEvent (cmd ref) resp] -> ([Operation (cmd ref) resp], Int)
```
Danten: Why return the `Int`? It seems you are just throwing it away.

PR #10, opened 2017-03-16
Use Property rather than Bool in the postcondition's codomain.

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -304,16 +305,25 @@ linearise
...
-    any' p xs = any p xs
+  go :: model -> [Rose (Operation cmd ix)] -> Property
+  go _ []       = property False
+  go m (o : os) = step m o .||. go m os
+
+  step :: model -> Rose (Operation cmd ix) -> Property
+  step m (Rose (Operation cmd resp _ _) roses) = postcondition m cmd resp .&&.
+    anyP (step (transition m cmd resp)) roses
+
+  anyP :: (a -> Property) -> [a] -> Property
```
Danten: I think we should rename this one since `anyP'` is the "real" `anyP`. Something like `anyPorDone`?

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -304,16 +305,25 @@ linearise
...
+linearise StateMachineModel {..} xs = go initialModel . linearTree $ xs
   where
-  postConditionsHold :: model -> Rose (Operation cmd ix) -> Bool
-  postConditionsHold m (Rose (Operation cmd resp _ _) roses) =
-    postcondition m cmd resp && any' (postConditionsHold (transition m cmd resp)) roses
-    where
-    any' :: (a -> Bool) -> [a] -> Bool
-    any' _ [] = True
-    any' p xs = any p xs
+  go :: model -> [Rose (Operation cmd ix)] -> Property
```
Danten: `go m = anyP' (step m)`

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -67,12 +67,12 @@ transitions m cmd _ = case cmd of
...
-postconditions :: Model -> MemStep resp  -> resp -> Bool
+postconditions :: Model -> MemStep resp  -> resp -> Property
 postconditions m cmd resp = case cmd of
-  New         -> True
-  Read ref    -> m  ! ref == resp && m' == m
-  Write ref i -> m' ! ref == i
-  Inc   ref   -> m' ! ref == m ! ref + 1
+  New         -> property $ True
+  Read ref    -> m  ! ref === resp .&&. m' == m
+  Write ref i -> property $ m' ! ref == i
```
Danten: couldn't this also be `m' ! ref === i`? I'm not sure about the combinators of `Property` but this looks inconsistent between `Read` and `Write`, `Inc`.

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -129,9 +128,19 @@ liftShrink' n0 shrinker = go n0
...
+shrinkPair :: (a -> [a]) -> (a, a) -> [(a, a)]
+shrinkPair shrinker = shrinkPair' shrinker shrinker
+
+shrinkPair' :: (a -> [a]) -> (b -> [b]) -> (a, b) -> [(a, b)]
+shrinkPair' shrinkerA shrinkerB (x, y) =
+  [ (x', y) | x' <- shrinkerA x ] ++
+  [ (x, y') | y' <- shrinkerB y ]
+
+------------------------------------------------------------------------
+
```
Danten: Couldn't we move these to some utility module?

src/Test/QuickCheck/StateMachineModel.hs
Danten: In fact if you define `go` that way, it might be better to inline it.

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: Yeah, I'm still playing around with the output. The `===` combinator adds a `counterexample` if the equality test fails, but they seem end up flooding the screen with newlines for some reason I don't get yet...

PR #21, opened 2017-03-24
Remove unsafeCoerce.

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -34,26 +33,38 @@ import           Text.PrettyPrint.ANSI.Leijen            (Pretty, align, dot,
...
+  :: forall cmd resp ix ref m
+  .  (Monad m, Enum ix, Typeable ref, RefKit cmd)
+  => Typeable resp
+  => Show resp
+  => (cmd resp ref -> m resp)
+  -> cmd resp ix -> StateT [ref] m resp
 liftSem sem cmd = do
 
   env <- get
   Untyped cmd' <- return $ fmap (\ix -> env !! fromEnum ix) $ Untyped cmd
```
Danten: So we have a `Functor (MemStepF resp)` instance but we don't use it?

stevana: I tried using it, but couldn't get it to work...

PR #22, opened 2017-03-30
Refactor/model parametrised by refs2

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -38,23 +40,25 @@ import           Test.QuickCheck.StateMachineModel.Utils
...
+
+
 liftSem
   :: forall cmd resp ix ref m
   .  (Monad m, Enum ix, Typeable ref, RefKit cmd)
   => Typeable resp
   => Show resp
   => (cmd resp ref -> m resp)
-  -> cmd resp ix -> StateT [ref] m resp
+  -> cmd resp ix -> StateT (Map Int ref) m resp
```
Danten: Why is this `Int` rather than `ix`? because you need to compute the next var? Why not take in a function:
`new :: Map ix ref -> ix`, and use that instead of `M.size`?

Danten: Hmm wait, now I'm getting confused again about what this is..

PR #23, opened 2017-03-31
Refactor/model parametrised by refs2 linearise

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -36,26 +40,58 @@ import           Text.PrettyPrint.ANSI.Leijen            (Pretty, align, dot,
...
+  Untyped cmd' <- return $ fmap (\ix -> env M.! ix) $ Untyped cmd
   let Just (cmd'' :: cmd resp ref) = ccast cmd'
 
   case returnsRef' cmd'' of
     Just Refl -> do
       ref <- lift $ sem cmd''
-      modify (++ [ ref ])
-      return ref
+      let ix :: ix
+          ix = toEnum . length . filter ((== pid) . snd) . M.keys $ env
```
Danten: Maybe having the variables not start from zero again would be a better idea? Might need to do less shifts then.

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -192,30 +246,42 @@ instance Pretty a => Pretty (Fork a) where
...
   right       <- fst <$> liftGen gens n
-  return $ Fork left prefix right
+  return $ Fork
+    (map (fmap (fixPid n 1)) left)
+    (map (fmap (\ix -> (ix, 0))) prefix)
+    (map (fmap (fixPid n 2)) right)
+  where
+  fixPid :: Int -> Int -> ix -> (ix, Int)
+  fixPid n pid ix | fromEnum ix < n = (ix, 0)
+                | otherwise       = (toEnum (fromEnum ix - n), pid)
```
Danten: indentation

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -433,12 +524,34 @@ class (Functor (Untyped cmd), Foldable (Untyped cmd)) => RefKit cmd where
...
     r = toEnum n
 
+  fixRefsPid
+    :: forall ix
+    .  (Enum ix, Ord ix)
+    => Int -> Int -> Untyped cmd (ix, Int) -> [Untyped cmd (ix, Int)] -> [Untyped cmd (ix, Int)]
+  fixRefsPid n pid c cs
+    | returnsRef c
+        = map (fmap (\(ix, pid') -> if pid == pid' && toEnum n < ix
+                                    then (toEnum (fromEnum ix - 1), pid')
```
Danten: `(pred ix, pid')`

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -34,72 +36,54 @@ newtype Ref = Ref Int
...
 
-transitions :: Model -> MemStep resp -> resp -> Model
-transitions m cmd _ = case cmd of
-  New         -> m ++ [0]
-  Read  _     -> m
-  Write ref i -> update m (unRef ref) i
-  Inc   ref   -> update m (unRef ref) ((m ! ref) + 1)
+transitions :: (Enum ref, Ord ref) => Model ref -> MemStep resp ref -> resp -> Model ref
+transitions (Model m) cmd ix = case cmd of
+  New         -> Model (M.insert ix 0 m) -- (toEnum $ length $ M.keys m) 0 m)
```
Danten: remove comment

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -129,23 +113,23 @@ debugMem ms0 = do
     v <- readIORef ref
     putStrLn $ "$" ++ show i ++ ": " ++ show v
   where
-  semSteps :: MonadIO io => [Untyped MemStepF Ref] -> io [IORef Int]
-  semSteps = flip execStateT [] . go
+  semSteps :: MonadIO io => [Untyped MemStep Ref] -> io [IORef Int]
+  semSteps = fmap (map snd . M.toList) . flip execStateT M.empty . go
```
Danten: `fmap M.elems`

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -38,28 +40,58 @@ import           Text.PrettyPrint.ANSI.Leijen            (Pretty, align, dot,
 
 import           Test.QuickCheck.StateMachineModel.Utils
 
+import           Unsafe.Coerce
+
 ------------------------------------------------------------------------
 
+liftSemSeq
```
stevana: Try reusing the liftSem involving pids for the sequential case also...

stevana:
Fixed things pointed out in the comments, thanks!

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -36,26 +40,33 @@ import           Text.PrettyPrint.ANSI.Leijen            (Pretty, align, dot,
...
 liftSem
   :: forall cmd resp ix ref m
   .  (Monad m, Enum ix, Typeable ref, RefKit cmd)
   => Typeable resp
+  => Ord (ix, Int)
   => Show resp
   => (cmd resp ref -> m resp)
-  -> cmd resp ix -> StateT [ref] m resp
-liftSem sem cmd = do
+  -> Int
```
Danten: Could we introduce an
```
type Pid = Int
```
Even if we don't expose it to the user it sure is nice to have some documentation, a `newtype` could also work.

stevana: Yep, I wanted to do this next. :-)

Probably something like CommandWithPid cmd, so we can avoid the Enum instance for pairs...

PR #25, opened 2017-04-04
Refactor/pid newtype

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -44,15 +43,18 @@ import           Unsafe.Coerce
...
 
+newtype Pid = Pid Int
+  deriving (Eq, Ord, Show, Read, Num)
+
 liftSem
   :: forall cmd resp ix ref m
   .  (Monad m, Enum ix, Typeable ref, RefKit cmd)
   => Typeable resp
-  => Ord (ix, Int)
+  => Ord (ix, Pid)
```
Danten: `Ord ix` is enough

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -326,38 +328,18 @@ linearTree es =
...
-  fromEnum (a, b) = cantorsPairing (fromEnum a) (fromEnum b)
-    where
-    cantorsPairing :: Int -> Int -> Int
-    cantorsPairing m n = (m + n) * (m + n + 1) `div` 2 + n
-
 linearise
   :: forall cmd ix model
-  .  (Enum ix, Ord (ix, Int))
-  => Show (model (ix, Int))
+  .  (Enum ix, Ord (ix, Pid))
```
Danten: `Ord ix`

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -104,7 +104,7 @@ semStep prb (Inc ref)     = liftIO $ do
 
 ------------------------------------------------------------------------
 
-debugMem :: MonadIO io => [Untyped MemStep (Ref, Int)] -> io ()
+debugMem :: MonadIO io => [Untyped MemStep (Ref, Pid)] -> io ()
```
Danten: Hmm I would like to move this file out from the example, make it take `semStep` as a parameter. Probably not in this PR, but we should really start to split up our files, and have the example only contain code that is specific for the example (the shorter the better).

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -217,7 +217,7 @@ prop_parallel prb = parallelProperty
 
 scopeCheck
```
Danten: Similarly this should also move out together with `debugMem` I'm not sure what to do about the properties that follow though. It would be nice to make sure that client code doesn't even know about `Pid` (or `Ref` for that matter, or even further it would be nice if the client doesn't even know that a program is `[Untyped cmd XXX]` 

So I don't think we can completely hide `Untyped` from the client, but it should probably only be used for `gens` and possibly `shrink1`

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: Yeah, perhaps `debugMem` can be generalised and grouped with the other debugging functions (`ppFork`, `debugShrink`, ...) at the bottom of the other module. Created #26.

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: Agreed, it would be nice not to leak `Pid`... Created #27.

PR #29, opened 2017-04-06
Refactor/internal ref instead of ix

stevana:


src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -202,37 +193,35 @@ prop_safety :: Problem -> Property
...
-      | returnsRef c = (s + 1, (toEnum s, pid) : refs)
+      | returnsRef c = (s + 1, IntRef (toEnum s) pid : refs)
       | otherwise    = (s,                       refs)
 
 prop_genScope :: Property
 prop_genScope = forAll (liftGenFork gens) $ \(Fork l p r) ->
   let p' = zip (repeat 0) p in
   scopeCheck (p' ++ zip (repeat 1) l) &&
-  scopeCheck (p' ++ zip (repeat 2) (r :: [Untyped MemStep (Ref, Pid)]))
+  scopeCheck (p' ++ zip (repeat 2) (r :: [Untyped MemStep IntRef]))
```
Danten: Is this type annotation still needed?

stevana: Nope, thanks!

PR #33, opened 2017-05-03
Refactor/single ref to multiple refs

stevana:


src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -52,147 +81,224 @@ newtype Ref = Ref Int
...
+    f resp refs -> MayResponse_ (ConstSym1 IntRef) resp -> Untyped' f refs
+
+data Operation cmd refs = forall resp.
+  (Show (Response_ refs resp),
+   Typeable resp,
+   Typeable (Response_ refs resp)) =>
+  Operation (cmd resp refs) (Response_ refs resp) Pid
+
+
+data Pair a :: (TyFun i *) -> *
```
Danten: This type is no longer used.

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -52,147 +81,224 @@ newtype Ref = Ref Int
...
+  go (c : cs) =
     [ [] ] ++
-    [ fixRefs n pid c cs ] ++
-    [ c' : cs' | (c', cs') <- shrinkPair' shrinker (go n') (c, cs) ]
-    where
-    n' = if returnsRef c then n + 1 else n
+    [ fixRefs c cs returns ] ++
+    [ c' : cs' | (c', cs') <- shrinkPair' shrinker go (c, cs) ]
+
+fixRefs
```
Danten: `fixRefs` doesn't so much fixes the refs but removes commands, rename?

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -215,39 +321,55 @@ instance Pretty a => Pretty (Fork a) where
...
     prefix
-    (map (fmap (fixPid n)) right)
+    (map (\(Untyped' cmd miref) ->
+            Untyped' (ifmap1 (fixPid ns) cmd) miref) right)
   where
-  fixPid :: Int -> IntRef -> IntRef
-  fixPid n (IntRef ix pid)
-    | fromEnum ix < n = IntRef ix 0
-    | otherwise       = IntRef ix pid
+  fixPid :: Map ix Int -> Sing (i :: ix) -> IntRef -> IntRef
```
Danten: Oh yeah, we should file an issue about fixing this (i.e the need for this).

src/Test/QuickCheck/StateMachineModel.hs
```diff
@@ -349,132 +466,213 @@ linearise StateMachineModel {..} xs0 = anyP (step initialModel) . linearTree $ x
...
-  :: forall cmd ref model
-  .  (Ord (Untyped cmd IntRef), Show (Untyped cmd IntRef))
-  => Typeable ref
-  => Show (model IntRef)
-  => RefKit cmd
+  :: forall
+     (ix    :: *)
+     (cmd   :: Response ix -> (TyFun ix * -> *) -> *)
+     (refs  :: TyFun ix * -> *)
+     (model :: (TyFun ix * -> *) -> *)
```
Danten: Is there a reason why these are not `ix ~> *` instead of `TyFun ix * -> *`?

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -95,133 +124,225 @@ semStep prb (Inc ref)     = liftIO $ do
...
-  fmap f (Inc   ref)   = Inc   (f ref)
-
-deriving instance Functor  (Untyped MemStep)
-deriving instance Foldable (Untyped MemStep)
+instance Read (Untyped' MemStep (ConstSym1 IntRef)) where
+  readPrec = parens $ choice
+    [ Untyped' <$ key "Untyped'" <*> parens (New <$ key " New") <*> readPrec
+    , Untyped' <$ key "Untyped'" <*> parens (Read <$ key "Read" <*> readPrec) <*> readPrec
+    , Untyped' <$ key "Untyped'" <*> parens (Write <$ key "Write" <*> readPrec <*> readPrec) <*> readPrec
+    , Untyped' <$ key "Untyped'" <*> parens (Inc <$ key "Inc" <*> readPrec) <*> readPrec
```
Danten: Hmm we should probably also have the `Copy` constructor here

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -95,133 +124,225 @@ semStep prb (Inc ref)     = liftIO $ do
...
-  show (Untyped c) = show c
+shrink1 :: Untyped' MemStep refs -> [Untyped' MemStep refs ]
+shrink1 (Untyped' (Write ref i) iref) = [ Untyped' (Write ref i') iref | i' <- shrink i ]
+shrink1 _                               = []
 
--- https://ghc.haskell.org/trac/ghc/ticket/8128
-instance (Read ref, Typeable ref, Show ref) => Read (Untyped MemStep ref) where
+------------------------------------------------------------------------
 
+instance Read (Untyped MemStep (ConstSym1 IntRef)) where
```
Danten: Is this one used?

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -95,133 +124,225 @@ semStep prb (Inc ref)     = liftIO $ do
...
+  showCmd (Inc   ref)   =  "Inc ("   ++ show ref ++ ")"
+  showCmd (Copy  ref)   =  "Copy ("   ++ show ref ++ ")"
+
+instance Eq (Untyped' MemStep (ConstSym1 IntRef)) where
+  Untyped' c1 _ == Untyped' c2 _ = Just c1 == cast c2
+
+instance Ord (Untyped' MemStep (ConstSym1 IntRef)) where
+  Untyped' c1 _ <= Untyped' c2 _ = Just c1 <= cast c2
+
+data RawMemStep refs
```
Danten: Do we use this for anything?

src/Test/QuickCheck/StateMachineModel/Example.hs
```diff
@@ -95,133 +124,225 @@ semStep prb (Inc ref)     = liftIO $ do
...
-scopeCheck :: RefKit cmd => [(Pid, Untyped cmd IntRef)] -> Bool
-scopeCheck = go 0 []
+usesRefs :: MemStep resp refs -> [Ex refs]
+usesRefs New           = []
+usesRefs (Read  ref)   = [Ex STuple0 ref]
+usesRefs (Write ref _) = [Ex STuple0 ref]
+usesRefs (Inc   ref)   = [Ex STuple0 ref]
+usesRefs (Copy  ref)   = [Ex STuple0 ref]
+
+scopeCheck
```
Danten: We should move this one, but we don't need to do it in this PR

src/Test/QuickCheck/StateMachineModel.hs
stevana: `stylish-haskell` doesn't seem to like `ix ~> *`, but it could be because of `TypeOperators` not being enabled; I'll try!

src/Test/QuickCheck/StateMachineModel.hs
stevana: Fixed, thanks!

src/Test/QuickCheck/StateMachineModel.hs
stevana: Filed #34.

src/Test/QuickCheck/StateMachineModel.hs
stevana: Fixed!

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: Fixed!

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: We should probably remove `Copy` and `Write`, as they make the minimal cases more complicated...

src/Test/QuickCheck/StateMachineModel/Example.hs
stevana: Yeah, needed for the above `Ord` instance.

src/Test/QuickCheck/StateMachineModel.hs
stevana: Doesn't work!

PR #36, opened 2017-05-05
Refactor/examples

stevana:


example/src/DieHard.hs
```diff
@@ -96,6 +99,60 @@ prop_dieSafety = sequentialProperty
...
+    , SmallIntoBig
+    , EmptyBig
+    , SmallIntoBig
+    , FillSmall
+    , SmallIntoBig
+    ]
+  ]
+
+testValidSolutions :: [Bool]
+testValidSolutions = map ((/= 4) . bigJug . run) validSolutions
```
Danten: Since this is not used anywhere how about:
```
testValidSolutions :: Bool
testValidSolutions = all ((== 4) . bigJug . run) validSolutions
```

example/src/DieHard.hs
```diff
@@ -96,6 +99,60 @@ prop_dieSafety = sequentialProperty
...
+  where
+  run = foldr (\c s -> transitions s c ()) initState
+
+prop_bigJug4 :: Property
```
Danten: weird number to start on

PR #40, opened 2017-05-05
Run `stack test` inside the example directory.

stevana:


stevana:
Not sure why this doesn't pass, I've reran it many times. Ideas?

stevana:
Ready for review!

PR #48, opened 2017-05-11
Simplify some of the types in the sequential module.y

stevana:


src/Test/StateMachine.hs
```diff
@@ -47,7 +47,7 @@ sequentialProperty
      (model :: (TyFun ix * -> *) -> *)
      (m     :: * -> *)
   .  Monad m
-  => Show (Untyped' cmd ConstIntRef)
+  => Show (IntRefed cmd)
```
Danten: Actually can we remove the kind signatures on this `forall`?

stevana: We could, I left it because I thought perhaps we can do better -- if we lift `go` to the top-level analogue to `checkParallelInvariant` below?

Danten: Ah that makes sense.

PR #53, opened 2017-05-12
Move returns into a type class.

stevana:


src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -218,36 +220,40 @@ mkHistoryKit pid = do
...
     atomically $ writeTChan (getHistoryChannel kit) $
       ResponseEvent (toDyn resp) (getProcessIdHistory kit)
 
 liftSemFork
-  :: SDecide ix
-  => IxFunctor cmd
+  :: forall
+     (ix    :: *)
+     (cmd   :: Response ix -> (TyFun ix * -> *) -> *)
+     (refs  :: TyFun ix * -> *)
```
Danten: Is this `forall` necessary?

example/src/MutableReference.hs
```diff
@@ -121,12 +120,12 @@ gens =
...
-returns (Read  _)   = SResponse
-returns (Write _ _) = SResponse
-returns (Inc   _)   = SResponse
-returns (Copy _)    = SReference STuple0
+instance HasResponse MemStep where
+  response New         = SReference STuple0
+  response (Read  _)   = SResponse
+  response (Write _ _) = SResponse
+  response (Inc   _)   = SResponse
+  response (Copy _)    = SReference STuple0
```
Danten: This could be:
```
instance HasResponse MemStep where
  response New   {} = SReference STuple0
  response Read  {} = SResponse
  response Write {} = SResponse
  response Inc   {} = SResponse
  response Copy  {} = SReference STuple0
```

stevana: Fixed, thanks!

src/Test/StateMachine/Internal/Parallel.hs
stevana: Seems so, `Could not deduce (SDecide ix0)` otherwise.

PR #57, opened 2017-05-15
Use Data.Tree instead of Rose.

stevana:


example/test/MutableReferenceSpec.hs
```diff
@@ -115,13 +116,13 @@ prop_shrinkForkMinimal = shrinkPropertyHelper (prop_parallel RaceCondition) $ \o
     = anyRose isMinimal
     . rose (liftShrinkFork shrink1)
     where
-    anyRose :: (a -> Bool) -> Rose a -> Bool
-    anyRose p (Rose x xs) = p x || any (anyRose p) xs
+    anyRose :: (a -> Bool) -> Tree a -> Bool
+    anyRose p (Node x xs) = p x || any (anyRose p) xs
```
Danten: `anyRose p = foldTree (\x y -> p x || or y)`

example/test/MutableReferenceSpec.hs
```diff
@@ -115,13 +116,13 @@ prop_shrinkForkMinimal = shrinkPropertyHelper (prop_parallel RaceCondition) $ \o
...
 
-    rose :: (a -> [a]) -> a -> Rose a
+    rose :: (a -> [a]) -> a -> Tree a
```
Danten: `rose f = unfoldTree (id &&& f)`

stevana: Fixed, thanks!

example/test/MutableReferenceSpec.hs
stevana: `foldTree` isn't part of `Data.Tree` in the `containers` version we are using... 

(Also an unfold followed by a fold is suggest we really need `hyloTree` :-).

PR #61, opened 2017-05-16
Factor out invariant checking and stats collecting from seq prop.

stevana:


src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -152,3 +158,39 @@ liftSem sem cmd iref = do
       ref <- lift $ sem cmd'
       modify $ IxM.insert i iref ref
       return iref
+
+------------------------------------------------------------------------
+
+collectStats :: [IntRefed cmd] -> Property -> Property
```
Danten: Could be:
```
collectStats :: [a] -> Property -> Property
```

PR #63, opened 2017-05-16
Make IxFunctor an endo functor.

stevana:


example/test/DieHardSpec.hs
```diff
@@ -1,3 +1,5 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
```
Danten: Why are we getting this warning now?

stevana: Because I didn't run `stack test` with the `--pedantic` flag earlier.

PR #65, opened 2017-05-16
Fix hlint hints.

stevana:


src/Test/StateMachine/Utils.hs
```diff
@@ -51,10 +50,11 @@ forAllShrinkShow gen shrinker shower pf =
 forAllShow
   :: Testable prop
   => Gen a -> (a -> String) -> (a -> prop) -> Property
-forAllShow gen shower pf = forAllShrinkShow gen (\_ -> []) shower pf
+forAllShow gen = forAllShrinkShow gen (const [])
```
Danten: Oh so when `hlint` says it you change it...

src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -150,7 +150,7 @@ liftSem
 liftSem sem cmd iref = do
 
   env <- get
-  let cmd' = ifmap @_ @_ @_ @_ @refs (\s i -> env IxM.! (s, i)) cmd
+  let cmd' = ifmap @_ @_ @_ @_ @refs (curry ((IxM.!) env)) cmd
```
Danten: I'm not sure I agree with this one.

src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -117,8 +118,7 @@ liftShrinkFork shrinker f@(Fork l0 p0 r0) =
       r'' = removeCommands p r
 
       removeManyCommands :: [IntRefed cmd] -> [IntRefed cmd] -> [IntRefed cmd]
-      removeManyCommands []       ds = ds
-      removeManyCommands (c : cs) ds = removeManyCommands cs (removeCommands c ds)
+      removeManyCommands cs ds = foldl (flip removeCommands) ds cs
```
Danten: I think it is easier to see what the recursive functions does.

src/Test/StateMachine/Utils.hs
stevana: ;-)

src/Test/StateMachine/Internal/Sequential.hs
stevana: Me neither, but then we need to figure out how to disable these kind of hints...

src/Test/StateMachine/Internal/Parallel.hs
stevana: Yeah...

stevana:
I've reverted the two hints that produced less readable code.

PR #71, opened 2017-05-17
Refactor/make alphaeq and scopecheck internal

stevana:


src/Test/StateMachine/Internal/ScopeCheck.hs
```diff
@@ -0,0 +1,38 @@
...
+scopeCheck
+  :: forall cmd. (IxFoldable cmd, HasResponse cmd)
+  => [(Pid, IntRefed cmd)] -> Bool
+scopeCheck = go []
+  where
+  go :: [IntRef] -> [(Pid, IntRefed cmd)] -> Bool
+  go _    []                           = True
+  go refs ((_, IntRefed c miref) : cs) = case response c of
+    SReference _  ->
+      let refs' = miref : refs in
```
Danten: Should we inline `refs'`?

src/Test/StateMachine/Internal/ScopeCheck.hs
```diff
@@ -0,0 +1,38 @@
...
+  ) where
+
+import           Test.StateMachine.Types
+import           Test.StateMachine.Utils
+
+------------------------------------------------------------------------
+
+scopeCheck
+  :: forall cmd. (IxFoldable cmd, HasResponse cmd)
+  => [(Pid, IntRefed cmd)] -> Bool
```
Danten: Maybe remove the `Pid` since it is not actually used?

src/Test/StateMachine/Internal/ScopeCheck.hs
stevana: Aye, thanks!

src/Test/StateMachine/Internal/ScopeCheck.hs
stevana: Fixed, thanks!

PR #74, opened 2017-05-18
Move frequency combinator from library to examples.

stevana:


example/src/DieHard.hs
```diff
@@ -64,8 +64,8 @@ smm = StateMachineModel preconditions postconditions transitions initState
 
 ------------------------------------------------------------------------
 
-gens :: [(Int, Gen (Untyped Step (RefPlaceholder ())))]
-gens =
+gen :: Gen (Untyped Step (RefPlaceholder ()))
+gen = frequency
```
Danten: Why not use `elements [Untyped FillBig, Untyped FillSmall, ...]`

stevana: Fixed, thanks!

PR #79, opened 2017-05-18
Fix hlint suggestions.

stevana:


src/Test/StateMachine/Types.hs
```diff
@@ -1,22 +1,20 @@
...
+{-# LANGUAGE IncoherentInstances       #-}
+{-# LANGUAGE MultiParamTypeClasses     #-}
+{-# LANGUAGE PolyKinds                 #-}
+{-# LANGUAGE Rank2Types                #-}
+{-# LANGUAGE ScopedTypeVariables       #-}
+{-# LANGUAGE TypeFamilies              #-}
+{-# LANGUAGE TypeInType                #-}
+{-# LANGUAGE TypeOperators             #-}
+{-# LANGUAGE UndecidableInstances      #-}
+{-# LANGUAGE UndecidableSuperClasses   #-}
```
Danten: So this is also annoying that `#-}` is indented based on the length on the longest names. Making it cumbersome to figure out exactly which extension have been removed (or maybe even added? who knows)

stevana: Yeah, I checked the options and there doesn't seem to be a fix...

PR #82, opened 2017-05-24
Feat/union find explict refs

stevana:


src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -110,6 +111,53 @@ liftGen gen pid ns = sized $ \sz -> runStateT (go sz) ns
...
+  runStateT (go cmds) ns
+  where
+
+  go :: [Untyped cmd (RefPlaceholder ix)] -> StateT (Map ix Int) Gen [IntRefed cmd]
+  go []                   = return []
+  go (Untyped cmd0 : cmds) = do
+
+    scopes <- get
+
+    Untyped cmd <- lift . genFromMaybe $ do
```
Danten: This will not work, if the cmd0 wants references that we don't have the this will always be return Nothing which will fail in genFromMaybe.

example/src/UnionFind.hs
```diff
@@ -0,0 +1,203 @@
...
+  n <- get
+  if n == 0
+  then do
+    modify succ
+    lift $ Untyped . New <$> arbitrary
+  else
+    lift $ frequency
+      [ (5, Untyped . Find <$> choose (0, n - 1))
+      , (5, Untyped <$> (Union <$> choose (0, n - 1) <*> choose (0, n - 1)))
+      ]
```
Danten: Doesn't this just generate one `New` and then random `Find` and `Union`?

stevana: Good catch, should be fixed now!

example/src/UnionFind.hs
```diff
@@ -0,0 +1,208 @@
...
+gen :: StateT Int Gen (Untyped Action (RefPlaceholder ()))
+gen = do
+  n <- get
+  if n == 0
+  then do
+    modify succ
+    lift $ Untyped . New <$> arbitrary
+  else do
+    cmd <- lift $ frequency
+      [ (1, Untyped . New  <$> arbitrary)
```
Danten: Need to `modify succ`

src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -75,40 +77,55 @@ liftGen
...
-      return . Just $ IntRef (Ref v) pid
+liftGen'
+  :: forall ix cmd s
+  .  Ord       ix
+  => SingKind  ix
+  => DemoteRep ix ~ ix
+  => IxTraversable cmd
+  => HasResponse cmd
+  => StateT s Gen (Untyped cmd (RefPlaceholder ix))
+  -> s
```
Danten: Could we pick a better name than `s` like `genState` or so, I'm not married to that name but this type is already quite big.

stevana: Fixed, thanks!

PR #86, opened 2017-05-29
Move mutable reference (meta)properties to separate module.

stevana:
(So that they can be imported and played around with in ghci.)


example/src/MutableReferenceProp.hs
```diff
@@ -0,0 +1,139 @@
+{-# OPTIONS_GHC -fno-warn-orphans #-}
+
+{-# LANGUAGE FlexibleInstances  #-}
+{-# LANGUAGE GADTs              #-}
+{-# LANGUAGE StandaloneDeriving #-}
+
+module MutableReferenceProp where
```
Danten: `MutableReference.Prop`?

stevana: Fixed, thanks!

PR #93, opened 2017-06-01
Add teaser to readme.

stevana:


README.md
```diff
@@ -8,13 +8,38 @@ stateful programs. The library is different from
...
+     New --> $0
+Parallel:
+  1. Inc $0 --> ()
+  2. Inc $0 --> ()
+     Read $0 --> 1
+```
+
+Clearly, if we increment a mutable reference in parallel we can end up with a
+race condition. We shall come back to this example, and link to the full source
+code, later.
```
Danten: Why tease a link? Write something like
> We shall come back to this example, but if your are impatient here is the [full source code example](https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/example/src/MutableReference.hs) .

stevana: Fixed, thanks!

PR #103, opened 2017-06-02
Try adding travis_retry to the test command.

stevana:


stevana:
Looking at the logs it seems that this actually does recovers from timeouts!

PR #104, opened 2017-06-05
Use BoxDrawings for parallel case

Danten:
Example output:

```
> quickCheck $ MutableReference.prop_parallel RaceCondition 
*** Failed! Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable (after 5 tests and 6 shrinks): 
Fork [Inc (IntRef (Ref 0) (Pid 0)) ()] [New (IntRef (Ref 0) (Pid 0))] [Inc (IntRef (Ref 0) (Pid 0)) (),Read (IntRef (Ref 0) (Pid 0)) ()]
Couldn't linearise:

Prefix: 
     New --> $0
Parallel:
              │ ┌────────┐
              │ │Inc $0  │
   ┌────────┐ │ │        │
   │Inc $0  │ │ │        │
   │  --> ()│ │ │        │
   └────────┘ │ │        │
              │ │  --> ()│
              │ └────────┘
              │ ┌────────┐
              │ │Read $0 │
              │ │   --> 1│
              │ └────────┘
```

new output
```
> quickCheck $ MutableReference.prop_parallel RaceCondition 
*** Failed! Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable, Falsifiable (after 4 tests and 4 shrinks): 
Fork [Inc (IntRef (Ref 0) (Pid 0)) ()] [New (IntRef (Ref 0) (Pid 0))] [Inc (IntRef (Ref 0) (Pid 0)) (),Read (IntRef (Ref 0) (Pid 0)) ()]
Couldn't linearise:

┌──────────────────────┐
│ New                  │
│                 ⟶ $0 │
└──────────────────────┘
           │ ┌─────────┐
           │ │ Inc $0  │
┌────────┐ │ │         │
│ Inc $0 │ │ │         │
│   ⟶ () │ │ │         │
└────────┘ │ │         │
           │ │    ⟶ () │
           │ └─────────┘
           │ ┌─────────┐
           │ │ Read $0 │
           │ │     ⟶ 1 │
           │ └─────────┘
```

### DIFF
--- a/src/Test/StateMachine/Internal/Parallel.hs
+++ b/src/Test/StateMachine/Internal/Parallel.hs
@@ -52,7 +52,7 @@ import           Data.Kind
                    (Type)
 import           Data.List
                    (partition)
-import qualified Data.Map                                as M
+import qualified Data.Map                                   as M
 import           Data.Maybe
                    (fromMaybe)
 import           Data.Singletons.Decide
@@ -70,16 +70,17 @@ import           Test.QuickCheck
 import           Test.QuickCheck.Monadic
                    (PropertyM)
 import           Text.PrettyPrint.ANSI.Leijen
-                   (Pretty, pretty, prettyList, text, vsep, (<+>))
+                   (Doc, Pretty, pretty, prettyList, text, vsep, (<+>))
 
 import           Test.StateMachine.Internal.IxMap
                    (IxMap)
-import qualified Test.StateMachine.Internal.IxMap        as IxM
+import qualified Test.StateMachine.Internal.IxMap           as IxM
 import           Test.StateMachine.Internal.Sequential
 import           Test.StateMachine.Internal.Types
 import           Test.StateMachine.Internal.Types.IntRef
                    (showRef)
 import           Test.StateMachine.Internal.Utils
+import           Test.StateMachine.Internal.Utils.BoxDrawer
 import           Test.StateMachine.Types
 
 ------------------------------------------------------------------------
@@ -254,6 +255,21 @@ toForkOfOps h = Fork (mkOps l) p' (mkOps r)
     dynResp = fromMaybe (error "toForkOfOps: impossible.") . fromDynamic
   mkOps _  = error "mkOps: Impossible."
 
+toBoxDrawings :: forall cmd. (IxFunctor cmd, ShowCmd cmd, HasResponse cmd) => History cmd -> Doc
+toBoxDrawings h = exec evT (fmap out (toForkOfOps h))
+  where
+    out :: [Operation cmd] -> [String]
+    out [] = []
+    out (Operation cmd resp _ : os) = showCmd (ifmap (const showRef) cmd)
+                                  : showResponse_ (response cmd) resp
+                                  : out os
+    toEventType :: History cmd -> [(EventType, Pid)]
+    toEventType = map $ \e -> case e of
+      InvocationEvent _ pid -> (Open, pid)
+      ResponseEvent _ pid   -> (Close, pid)
+    evT :: [(EventType, Pid)]
+    evT = toEventType (filter (\e -> getProcessIdEvent e `elem` [1,2]) h)
+
 ------------------------------------------------------------------------
 
 data HistoryKit cmd refs = HistoryKit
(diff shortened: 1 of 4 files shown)


src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -254,6 +256,24 @@ toForkOfOps h = Fork (mkOps l) p' (mkOps r)
     dynResp = fromMaybe (error "toForkOfOps: impossible.") . fromDynamic
   mkOps _  = error "mkOps: Impossible."
 
+toBoxDrawings :: forall cmd. (IxFunctor cmd, ShowCmd cmd, HasResponse cmd) => History cmd -> Doc
```
stevana: Would it make sense to move this (or some of it?) into the box drawing module?

Maybe align and add type signatures to help readability?

src/Test/StateMachine/Internal/Utils/BoxDrawer.hs
```diff
@@ -0,0 +1,70 @@
+module Test.StateMachine.Internal.Utils.BoxDrawer
+  ( EventType(..)
+  , exec)
+  where
+
+import           Text.PrettyPrint.ANSI.Leijen
+                   (Doc, text, vsep)
+
+type Pid = Int
```
stevana: Can we reuse already existing types here?

src/Test/StateMachine/Internal/Utils/BoxDrawer.hs
```diff
@@ -0,0 +1,70 @@
...
+size Deactive  = 0
+size (Ret l)   = 10 + length l
+size Bottom    = 4
+
+adjust :: Int -> Cmd -> String
+adjust n Top = " ┌" ++ replicate (n - 4) '─' ++ "┐"
+adjust n (Start l) = " │" ++ l ++ replicate (n - length l - 5) ' ' ++ " │"
+adjust n Active = " │" ++ replicate (n - 4) ' ' ++ "│"
+adjust n Deactive = replicate (n - 1) ' '
+adjust n (Ret l) = " │" ++ replicate (n - 8 - length l) ' ' ++ "--> " ++ l ++ "│"
```
stevana: Might as well use \longarrow here?

stevana:
It would be nice if the prefix was also "boxed". 

I think ideally we should be able to drop the "Prefix" and "Parallel" text from the output and it should still be clear to the user what's going on...

Also: Module header missing, haddock comments missing, and readme needs updated.

src/Test/StateMachine/Internal/Utils/BoxDrawer.hs
stevana: We should perhaps also open a ticket on making the utf8 output optional (or even better, detect if utf8 is supported by the terminal, if not use ascii)?

src/Test/StateMachine/Internal/Parallel.hs
Danten: `History` and `Operation` are defined in this module so we would get a cyclic import if I move stuff that depends on that from here.

stevana: I put some stuff in https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/src/Test/StateMachine/Internal/Types/IntRef.hs to avoid cycling imports, perhaps you can do a similar thing?

Danten: I think I will leave that to a later refactoring

src/Test/StateMachine/Internal/Utils/BoxDrawer.hs
```diff
@@ -0,0 +1,97 @@
...
+-- |
+-- Module      :  Test.StateMachine.Internal.Parallel
+-- Copyright   :  (C) 2017, ATS Advanced Telematic Systems GmbH
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  Stevan Andjelkovic <stevan@advancedtelematic.com>
+-- Stability   :  provisional
+-- Portability :  non-portable (GHC extensions)
+--
+-- This module contains the building blocks needed to implement the
```
stevana: Copy paste...

src/Test/StateMachine/Internal/Utils/BoxDrawer.hs
```diff
@@ -0,0 +1,97 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Test.StateMachine.Internal.Parallel
+-- Copyright   :  (C) 2017, ATS Advanced Telematic Systems GmbH
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  Stevan Andjelkovic <stevan@advancedtelematic.com>
```
stevana: This is your module!

PR #106, opened 2017-06-06
Feat/comment union find

stevana:


src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -255,7 +255,8 @@ checkSequentialInvariant
     monitor $ label s
     pre $ precondition m cmd
     resp <- run $ liftSem (sem m miref) cmd miref
+    let m' = transition m cmd resp
     liftProperty $
-      counterexample ("\nThe post-condition for `" ++ s ++ "' failed!\n\n" ++ show m) $
+      counterexample ("\nThe post-condition for `" ++ s ++ "' failed!\n\n" ++ show m') $
```
Danten: Hmm? This seems weird to print only the transitioned one.

stevana: I basically just did it this way to get the desired output for the die hard example.

You reckon we should print both? Something like: "The model was `m`, then `s` happend, post-condition for `s` doesn't hold in the new model `m'`?

Danten: But the post-condition should hold in the **old** model, not the new.

stevana: I see what you mean with it being weird to print the transitioned one now, :-).

Hmm...

Danten: So I still think we can print both, but we need to word it correctly.

PR #107, opened 2017-06-06
Comment the mutable reference example.

stevana:


example/src/MutableReference.hs
```diff
@@ -40,6 +54,12 @@ import           Test.StateMachine
 
 ------------------------------------------------------------------------
 
+-- Mutable references can be created, read from, written to and
+-- incremented. Note that the @New@ action returns a @Reference '()@,
+-- and all other actions take references, @refs \@\@ '()@, as arguments
+-- -- these are special library constructs that help you work with
```
Danten: double `--`

stevana: I always use double -- -- not sure why!

PR #121, opened 2017-06-22
Respect pre-conditions while generating and shrinking.

stevana:


example/src/UnionFind.hs
```diff
@@ -175,23 +175,26 @@ smm = StateMachineModel preconditions postconditions transitions initModel
...
+     ]
+  , gprecondition = gprecondition'
+  , gtransition   = gtransition'
+  , initGenState  = 0
+  }
+  where
+  gprecondition' :: Int -> Action refs resp -> Bool
+  gprecondition' _ (New   _)   = True
+  gprecondition' n (Find  _)   = n > 0
+  gprecondition' n (Union _ _) = n > 0
```
Danten: Shouldn't these be
```
gprecondition' n (Find  x)   = withIn n x
gprecondition' n (Union x y) = withIn n x

withIn n x = 0 <= x && x < n -- or some better name
```

stevana: Possibily, that's already taken care of by the `StateMachineModel`'s `precondition` though... We probably want to merge the two concepts somehow...

Danten: But shrinking doesn't use `StateMachineModel` but uses this one right?

stevana: Ah, yes!

PR #124, opened 2017-06-27
Quick and dirty fix for making the test pass on CI.

stevana:
The proper fix is to generalise the machinery in
`MutualReference.Prop.prop_shrinkForkMiminal` and use it in
`TicketDispenser.prop_parallelBad`.


stevana:
Lets solve this in one of the ways described in the above ticket instead.

PR #127, opened 2017-06-29
add shrinking

Danten:
So this shrink removes random commands, and then goes through and remove commands that uses vars that are no longer in scope or the precondition doesn't hold anymore for.

### DIFF
--- a/src/Test/StateMachine/Prototype.hs
+++ b/src/Test/StateMachine/Prototype.hs
@@ -25,10 +25,14 @@ import           Control.Monad.State
 import           Data.Dynamic
 import           Data.Functor.Classes
                    (Eq1(..), Ord1(..), Show1(..), showsPrec1)
+import           Data.Functor.Const (Const(..))
 import           Data.IORef
 import           Data.Map
                    (Map)
 import qualified Data.Map                            as M
+import           Data.Set
+                   (Set)
+import qualified Data.Set                            as Set
 import           Data.Tree
 import           System.Random
                    (randomRIO)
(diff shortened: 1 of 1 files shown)


src/Test/StateMachine/Prototype.hs
```diff
@@ -279,8 +290,41 @@ liftGen' gen pre next model0 n = sized $ \size -> go size n model0
...
+getUsedVars = hfoldMap (\(Symbolic v) -> Set.singleton v)
+
+filterPrecondition
+  :: HFoldable act
+  => Precondition model act
+  -> Transition model act
+  -> model Symbolic
+  -> Set Var
+  -> [Internal act] -> ((model Symbolic, Set Var), [Internal act])
+filterPrecondition pre trans = go
```
stevana: Better name? `filterInvalid`?

src/Test/StateMachine/Prototype.hs
```diff
@@ -450,10 +495,40 @@ liftGenFork gen pre next model = do
...
+  :: HFoldable act
+  => (forall v resp. act v resp -> [act v resp])
+  -> Precondition model act
+  -> Transition model act
+  -> model Symbolic
   -> (Fork [Internal act] -> [Fork [Internal act]])
-liftShrinkFork _ _ = []
+liftShrinkFork oldShrink pre trans model (Fork l p r) =
+  map (forkFilterPrecondition pre trans model)
+  [ Fork l' p' r' | (p', (l', r')) <- shrinkPair shrinkSub (shrinkPair shrinkSub shrinkSub) (p, (l, r))]
```
stevana: Would it be possible to inline the filtering somehow here?

Danten: I'm not sure that is beneficial, I will leave it for now like it is for the similarity with `liftShrink`

PR #128, opened 2017-06-29
fix bound var

Danten:
it was printed as `(Var i) ← New`, rather than `$i ← New`

```
Couldn't linearise:

┌───────────────────────────┐
│ $3 ← New                  │
│                  ⟶ Opaque │
└───────────────────────────┘
┌─────────┐ │                
│ Inc $3  │ │                
│         │ │ ┌─────────────┐
│         │ │ │ Write $3 -2 │
│         │ │ │        ⟶ () │
│         │ │ └─────────────┘
│    ⟶ () │ │                
└─────────┘ │                
┌─────────┐ │                
│ Read $3 │ │                
│     ⟶ 1 │ │                
└─────────┘ │                
```

### DIFF
--- a/src/Test/StateMachine/Prototype.hs
+++ b/src/Test/StateMachine/Prototype.hs
@@ -81,8 +81,11 @@ instance Ord1 Symbolic where
 
 newtype ShowSymbolic a = ShowSymbolic Var
 
+showVar :: Var -> String
+showVar (Var i) = "$" ++ show i
+
 instance Show1 ShowSymbolic where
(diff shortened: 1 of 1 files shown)


src/Test/StateMachine/Prototype.hs
```diff
@@ -81,8 +81,11 @@ instance Ord1 Symbolic where
 
 newtype ShowSymbolic a = ShowSymbolic Var
 
+showVar :: Var -> String
```
stevana: Perhaps we should use the `Pretty` type class for these kind of things?

PR #129, opened 2017-07-03
Refactor/simplify refs

stevana:


src/Test/StateMachine/Internal/Sequential.hs
```diff
@@ -26,255 +17,118 @@
...
-liftGenHelper Generator {..} pid ns = sized $ \sz ->
-  runStateT (runStateT (go sz) initGenState) ns
+  :: forall model act
+  .  Generator model act
+  -> Precondition model act
+  -> Transition model act
+  -> model Symbolic
+  -> Int
+  -> Gen ([Internal act], model Symbolic, Int)
+liftGenHelper gen precond next model0 n = sized $ \size -> go size n model0
```
Danten: The `Int` that is returned will be `n + 2*size`?

stevana: Removed the helper completely, thanks!

example/src/UnionFind.hs
```diff
@@ -110,149 +101,131 @@ instance Show a => Show (Element a) where
...
+    "Union (" ++ show ref1 ++ ") (" ++ show ref2 ++ ")"
 
 ------------------------------------------------------------------------
 
 -- The model corresponds closely to the *relational specification* given
 -- in section 12 of the paper.
 
-newtype Model refs = Model [Element Int]
-  deriving Show
+newtype Model a (v :: * -> *) = Model [(Ref a v, Ref a v)]
```
stevana: I'll go throught all comments next and update them if needed, thanks!

example/src/MutableReference.hs
```diff
@@ -24,209 +21,173 @@
...
+unRef :: Ref Concrete -> IORef Int
+unRef (Ref (Concrete (Opaque ref))) = ref
+
+instance Eq1 v => Eq (Ref v) where
+  Ref v1 == Ref v2 = liftEq (==) v1 v2
+
+instance Show1 v => Show (Ref v) where
+  show (Ref v) = showsPrec1 10 v ""
+
+instance Show (Internal Action) where
```
stevana: Yeah something like this would be nice; I'll keep it in mind as I refactor the types module (still not happy with it).

PR #132, opened 2017-07-04
One way of combining models out of smaller ones

Danten:
Mostly for fun


stevana:
I think we should try to make use of this in a "real" example, and also open an issue about composition of specs more generally!

PR #135, opened 2017-07-05
Try to clean up the show related code a bit.

stevana:


example/src/MutableReference.hs
```diff
@@ -71,20 +69,15 @@ instance Eq1 v => Eq (Ref v) where
...
-  show (Internal (Inc   ref)   sym) =
-    "Inc ("   ++ show ref ++ ") (" ++ show sym ++ ")"
-
 instance ShowAction Action where
-  showAction New           = ShowResponse "New"                                   show
-  showAction (Read ref)    = ShowResponse ("Read " ++ show ref)                   show
-  showAction (Write ref i) = ShowResponse ("Write " ++ show ref ++ " " ++ show i) show
-  showAction (Inc ref)     = ShowResponse ("Inc " ++ show ref)                    show
+  showAction New           =
+    ShowResponse "New"                                             show
```
Danten: Maybe we should add `showResponse` smart constructor that always uses `show` for the response.

stevana: Fixed, thanks!

PR #138, opened 2017-07-06
split out helper properties into more combinator style

Danten:
This change also exposes the types of `Fork` and `Program` but not its constructors, so at the moment one can only plug and play in a similar fashion as the property helpers anyway (unless one imports the `Internal`), but I think it is okay.

Maybe we should also expose `checkParallelInvariant`?

### DIFF
--- a/src/Test/StateMachine.hs
+++ b/src/Test/StateMachine.hs
@@ -17,71 +17,105 @@
 -----------------------------------------------------------------------------
 
 module Test.StateMachine
-  ( -- * Sequential property helper
-    sequentialProperty
-  , sequentialProperty'
-    -- * Parallel property helper
-  , parallelProperty
-  , parallelProperty'
+  ( -- * ForAll combinator for program
+    forAllProgram
+  , forAllParallelProgram
+    -- * Run sequential program, and check model
+  , runSequentialProgram
+  , runSequentialProgram'
+    -- * Run parallel program
+  , runParallelProgram
+  , runParallelProgram'
+  , checkParallelInvariant
   , module Test.StateMachine.Types
+  , ParallelProgram
+  , Program
   ) where
 
 import           Control.Monad.State
                    (evalStateT, replicateM_)
 import           Test.QuickCheck.Monadic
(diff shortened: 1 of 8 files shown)


stevana:
Nice!

Perhaps we should introduce `ParallelProgram act = Fork (Program act)`?

src/Test/StateMachine.hs
```diff
@@ -17,29 +17,150 @@
...
+  -> Transition model act
+  -> InitialModel model
+  -> (Program act -> Property)
+  -> Property
+forAllProgram gen shrinker precond trans m =
+  forAllShrink
+    (fst <$> liftGen gen precond trans m 0)
+    (liftShrink shrinker precond trans m)
+
+-- | This function is like a `forAllShrink` for `Fork (Program act)`,
```
stevana: Outdated comment.

src/Test/StateMachine.hs
```diff
@@ -17,29 +17,150 @@
...
+    monadic (ioProperty . runnerWithSetup)
+      (liftModel m m acts precond sem trans postcond)
+  where
+  runnerWithSetup mp = do
+    s <- setup
+    let prop = runner s (evalStateT mp emptyEnvironment)
+    cleanup s
+    return prop
+
+-- | Run a parallel program `Fork (Program act)` which provides
```
stevana: Outdated comment.

src/Test/StateMachine.hs
```diff
@@ -106,12 +217,13 @@ parallelProperty
...
   -> Transition    model act
   -> Postcondition model act
-  -> (forall v. model v)                                  -- ^ Initial model
+  -> InitialModel model      -- ^ Initial model
   -> Semantics act IO
   -> Property
 parallelProperty gen shrinker precond trans postcond initial sem =
-  parallelProperty' gen shrinker precond trans postcond
-    initial (return ()) (const sem) (const (return ()))
+  forAllForkProgram gen shrinker precond trans initial $ \fork ->
```
stevana: s/Fork/Parallel/

PR #139, opened 2017-07-10
Refactor/more docs

stevana:


src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -73,12 +69,135 @@ import           Test.StateMachine.Types
...
+  -> Precondition model act
+  -> Transition   model act
+  -> model Symbolic
+  -> Gen (ParallelProgram act)
+generateParallelProgram generator precondition transition model = do
+  let generate     =  generateProgram generator precondition transition
+  (prefix, model') <- runStateT  (generate 0) model
+  let offset       =  length (unProgram prefix) + 1
+  left             <- evalStateT (generate offset) model'
+  right            <- evalStateT (generate offset) model'
```
Danten: Wait a minute, shouldn't this be
```haskell
right <- evalStateT (generate $ offset + 1 + length (unProgram left)) model'
```
I'm not sure if the `1 +` is necessary...

stevana: Fixed in below commit, thanks!

example/test/MutableReferenceSpec.hs
```diff
@@ -18,8 +18,11 @@ spec = do
     it "`prop_genScope`: generate well-scoped programs"
       prop_genScope
 
-    it "`prop_genForkScope`: generate well-scoped parallel programs"
-      prop_genForkScope
+    it "`prop_genParallelScope`: generate well-scoped parallel programs"
+      prop_genParallelScope
+
+    it "`prop_genParallelSequence`: generate ..."
```
Danten: > generate unique names in order

Or something like that

PR #142, opened 2017-07-11
Refactor/readme

stevana:


README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
...
 
-Couldn't linearise:
+In order to be able to show counterexamples, we need a show instance for our
+actions. `IORef`s don't have a show instance, thats why we wrap them in
+`Opaque`; which gives a show instance to a type that doesn't have one.
+
+When we generate actions we won't be able to create arbitrary `IORef`s, that's
+why all uses of `IORefs` are wrapped in `Reference v`, where the parameter `v`
+will let us use symbolic references while generating (and concrete ones when
+executing).
```
Danten: I think it would be better to swap the order of these two paragraphs. I want the `Opaque` type to be more of a side-remark, but having it first makes it look like it is very important.

README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
 `quickcheck-state-machine` library can be seen as an attempt to provide similar
 functionality to Haskell's QuickCheck library.
 
-### Sample run (teaser)
+### Example
 
-Here's a sample output from when we look for race conditions in the mutable
-reference example:
+Lets implement and test programs using mutable references. Our mutable
```
Danten: > Lets implement and test programs using mutable references, which will be implemented using `IORef Int`.

README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
 `quickcheck-state-machine` library can be seen as an attempt to provide similar
 functionality to Haskell's QuickCheck library.
 
-### Sample run (teaser)
+### Example
 
-Here's a sample output from when we look for race conditions in the mutable
-reference example:
+Let's implement and test programs using mutable references. Our mutable
+references can be created, read from, written to and incremented:
```
Danten: Let's mplement and test programs using mutable references implemented using `IORef Int`.

README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
...
+passes. If we however introduce the bug problem, then it will fail with the
+minimal counterexample:
 
 ```
+> quickCheck (prop_references Bug)
+*** Failed! Falsifiable (after 16 tests and 4 shrinks):
+[New (Var 0),Write (Var 0) 5 (Var 2),Read (Var 0) (Var 3)]
+Just 5 /= Just 6
+```
+
```
Danten: Make a remark why this went wrong, the fact that we did a `Write` in the `[5 .. 10]` range which is "buggy".

README.md
Danten: Our mutable references can be created, read from, written to and incremented, which we will represent using the `Action` type:

README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
...
+│        ⟶ () │ │ │              │
+└─────────────┘ │ │              │
+                │ │         ⟶ () │
+                │ └──────────────┘
+                │ ┌──────────────┐
+                │ │ Read (Var 0) │
+                │ │          ⟶ 1 │
+                │ └──────────────┘
+Just 2 /= Just 1
+```
```
Danten: Quick comment how to read this diagram, that explains that this program creates a new `Var 0`, and then in parallel (concurrently) does two `Inc`, but somehow only reads a `1` afterwards. Which if one looks in the semantics can see that we have a race condition.

README.md
```diff
@@ -18,36 +18,203 @@ testing first appeard in Erlang's proprietary QuickCheck. The
...
+  initModel $ \prog ->
+    runAndCheckProgram
+      precondition
+      transition
+      postcondition
+      initModel
+      (semantics prb)
+      ioProperty
+      prog
+
```
Danten: Maybe move the parallel property definition below the running of the sequential, motivating race condition...

PR #145, opened 2017-09-05
Add CRUD webserver example.

stevana:


example/src/CrudWebserverDb.hs
```diff
@@ -0,0 +1,415 @@
...
+instance ToJSON User where
+  toJSON (User name age) =
+    object [ "name" .= name
+           , "age"  .= age
+           ]
+
+-- For testing we will also need to generate arbitrary users.
+
+instance Arbitrary User where
+  arbitrary = User <$> (T.pack <$> (arbitrary `suchThat` all isPrint))
```
Danten: Should maybe have this be
```haskell
arbitrary = User <$> (T.pack <$> (listOf (arbitrary `suchThat` isPrint)))
                 <*> arbitrary
```
To be more likely to succeed in generating longer names.

example/src/CrudWebserverDb.hs
```diff
@@ -0,0 +1,415 @@
...
+runServer :: FilePath -> Warp.Port -> IO ()
+runServer sqliteFile port = Warp.run port =<< mkApp sqliteFile
+
+------------------------------------------------------------------------
+
+-- We get the client bindings for free from the API datatype.
+
+postUserC    :: User     -> ClientM (Key User)
+getUserC     :: Key User -> ClientM (Maybe User)
+putUserC     :: Key User -> ClientM ()
```
Danten: `putUserC`? Seems like a strange name, shouldn't it be something like `incUserC` or even `incAgeUserC`.

example/src/CrudWebserverDb.hs
```diff
@@ -0,0 +1,415 @@
...
+-- start with the specification and testing using the
+-- <https://github.com/advancedtelematic/quickcheck-state-machine
+-- quickcheck-state-machine> library.
+
+-- We start by specifying what actions are allowed.
+
+data Action (v :: * -> *) :: * -> * where
+  PostUser   :: User                   -> Action v (Key User)
+  GetUser    :: Reference v (Key User) -> Action v (Maybe User)
+  PutUser    :: Reference v (Key User) -> Action v ()
```
Danten: Same as above, shouldn't this be called `Inc` something?

stevana:
Fixed, thanks!

PR #147, opened 2017-09-25
Refactor/forall

stevana:
Had a go at refactoring the combinators. Not entirely sure about the design choices yet.

If the post-condition would be boolean, we could have `runCommands` return a `Result = Ok | PreCondFailed | PostCondFailed...` and then do:

```haskell
(hist, model, res) <- runCommands prog
print hist `whenFail` res === Ok
```

Which would be closer to what they do in Erlang, https://github.com/Quviq/QuickCheckExamples/blob/master/src/crud_eqc.erl.

Also managed to unify the semantics/runners of sequential and parallel by using `lifted-async`, it could be argued that `MonadBaseControl` complicates things though...

What do you reckon?

### DIFF
--- a/src/Test/StateMachine.hs
+++ b/src/Test/StateMachine.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE DuplicateRecordFields #-}
 {-# LANGUAGE FlexibleContexts      #-}
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE Rank2Types            #-}
@@ -25,11 +26,13 @@ module Test.StateMachine
     Program
   , programLength
   , forAllProgram
-  , runAndCheckProgram
-  , runAndCheckProgram'
   , StateMachine(..)
+  , Generation(..)
+  , Execution(..)
   , monadicSequential
+  , monadicSequential'
   , runCommands
+  , runProgram
   , prettyCommands
   , commandNames
   , checkCommandNames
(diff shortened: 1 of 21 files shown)


stevana:
Ready for review!

(Decided to keep `Property` rather than `Bool` for post-conditions for now, and stick with `MonadBaseControl` and see how it goes.)

src/Test/StateMachine.hs
```diff
@@ -70,47 +90,51 @@ forAllProgram generator shrinker precondition transition model =
...
-    (checkProgram precond trans postcond m m sem acts)
+  -> PropertyM m (History act, model Concrete, Property)
+runProgram sm = run . executeProgram sm
+
+prettyProgram
+  :: MonadIO m
+  => Program act -> History act -> model Concrete -> Property -> PropertyM m ()
+prettyProgram _ hist _ prop = putStrLn (ppHistory hist) `whenFailM` prop
+
+actionNames :: forall act. Show (Untyped act) => Program act -> [(String, Int)]
```
Danten: Similar here we can use `Data.Data` to get the names of the constructors instead of using `Show`

```haskell
  name = show . toConstr
```

Danten: Of course that means that all examples will have to derive `Data`, and hmm that seems complicated for GADTs..

stevana: Yeah... I just tried and deriving `Data` fails. Same with `Generic` which would have been another option, so I guess TH it is?

Issue #77, opened 2017-05-18
Can we show results when printing sequential counterexamples?


stevana:
I.e. print `Operation`s rather than just commands?

stevana:
This can probably be achieved by changing `checkSequentialInvariant` to bookkeep a list of `Operation`s as it works its way though the commands.

stevana:
Running sequential programs also processes a `History` as of #147, the actual pretty printing of it can still be improved though.

stevana:
I had an idea of how we might want format the output. Basically inspired by Floyd-Hoare logic:
```
{}                                -- Initial model
    ref <- New              -- References "bind" their result.
{ ref = 0 }                    -- Model after the action.
    Write ref 2 --> ()     -- we might want to hide the "--> ()" for unit results.
{ ref = 2 }
    Read ref --> 2
{ ref = 2 }
...
```

We could perhaps use something like https://hackage.haskell.org/package/pretty-show to show the model in a nicely formatted way.

Thoughts?

stevana:
This is done as of #209, including diffs of models for better readability.

PR #155, opened 2017-09-29
Cosmetic changes

Lysxia:


stevana:
Actually, this seems to break some tests, e.g.:

```
DieHard, Sequential property, `prop_bigJug4`: in most cases, the smallest solution is found

       uncaught exception: ErrorCall (Prelude.read: no parse) (after 1 test)

       failed: *** Failed! Falsifiable (after 49 tests and 8 shrinks): 

       [FillSmall :> Var 8,SmallIntoBig :> Var 11,FillSmall :> Var 12,SmallIntoBig :> Var 20,EmptyBig :> Var 21,SmallIntoBig :> Var 22,FillSmall :> Var 23,SmallIntoBig :> Var 24]
```

The problem is that `Read` instances don't match up with your changes to `Show`.

I think we should leave the show instances as they are, i.e. they should be Haskell expressions in string form, and instead pretty print the history as suggested in #77. I've done this as part of  #147, see `ppHistory`.

Lysxia:
I agree that pretty-printing history seems to be a better approach.

`Show` would be used elsewhere perhaps. Still, the current `Show` doesn't quite produce valid expression either, it would look like:

    [FillSmall (Var 8), ...]

But the first value here is in fact a pair of `FillSmall` (wrapped in some stuff) and `Var 8`.

Perhaps `Read` is the one that also needs a fix?

stevana:
You are right, we should probably keep the `Internal` constructor when showing and change the read instances...

I've created issue #157 for this.

Lysxia:
I'll come back with a better fix.

PR #156, opened 2017-09-29
Add Lysxia to license file, update readme and changelog with circular

stevana:
buffer example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,11 @@
 #### ???
 
-  * Two examples of how to test CRUD web applications were added.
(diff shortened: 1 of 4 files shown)


Lysxia:
Awesome, thanks!

Just to let you know, and in case that matters, Li-yao is actually my first name. I just put my last name first when I write my full name sometimes.

stevana:
Got it! I'm happy with which ever order you feel the most comfortable with, let me know if you'd prefer the order swapped.

Lysxia:
It's fine like this for me!

PR #158, opened 2017-09-30
Integrate Quickcheck with counterexamples

### LINKED ISSUE #157: Show instances produce invalid expressions.
The `Show` instance for `Program` drops `Internal` and `Symbolic` constructors from the output, which produces invalid expressions, e.g.
```
[FillSmall (Var 8), ...]
```
Now that we pretty print the history of a run, instead of the program, we should fix the above and change the `Read` instances accordingly.

Lysxia:
- Add variants of properties which return counterexamples, so we don't need to read them from strings anymore.
- Add a type class to recognize constructors. A TH script to write the instances is TODO.
- Fix `Show` instances. TODO: This makes traces pretty noisy, looking like `Write (Reference (Symbolic (Var 0))) 0` instead of `Write (Var 0) 0`. But instead of hijacking `Show`, it seems better to specify a custom printer of actions separately, via an explicit parameter or a different type class.
- Remove `Read` instances in examples and unused imports.

Fixes #157

### DIFF
--- a/example/src/DieHard.hs
+++ b/example/src/DieHard.hs
@@ -30,6 +30,8 @@ import           Data.Functor.Classes
                    (Show1)
 import           Test.QuickCheck
                    (Property, elements, property)
+import           Test.QuickCheck.Counterexamples
+                   (PropertyOf)
 
 import           Test.StateMachine
 
@@ -171,8 +173,8 @@ sm = StateMachine
   generator shrinker preconditions transitions
   postconditions initModel semantics id
 
-prop_dieHard :: Property
-prop_dieHard = monadicSequential sm $ \prog -> do
+prop_dieHard :: PropertyOf (Program Action)
+prop_dieHard = monadicSequentialC sm $ \prog -> do
   (hist, model, prop) <- runProgram sm prog
   prettyProgram prog hist model $
     checkActionNames prog prop
--- a/example/src/DieHard.hs
+++ b/example/src/DieHard.hs
@@ -29,7 +29,7 @@ module DieHard
 import           Data.Functor.Classes
                    (Show1)
 import           Test.QuickCheck
-                   (Property, elements, property)
+                   (elements, property)
 import           Test.QuickCheck.Counterexamples
                    (PropertyOf)
 
--- a/example/src/DieHard.hs
+++ b/example/src/DieHard.hs
@@ -149,6 +149,16 @@ instance HTraversable Action where
   htraverse _ SmallIntoBig = pure SmallIntoBig
   htraverse _ BigIntoSmall = pure BigIntoSmall
 
+instance Constructors Action where
+  constructor x = Constructor $ case x of
+    FillBig      -> "FillBig"
+    FillSmall    -> "FillSmall"
+    EmptyBig     -> "EmptyBig"
+    EmptySmall   -> "EmptySmall"
+    SmallIntoBig -> "SmallIntoBig"
+    BigIntoSmall -> "BigIntoSmall"
+  nConstructors _ = 6
+
 ------------------------------------------------------------------------
 
 -- Finally we have all the pieces needed to get the sequential property!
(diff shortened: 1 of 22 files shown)


example/src/DieHard.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE GADTs              #-}
 {-# LANGUAGE KindSignatures     #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeOperators      #-}
```
stevana: Is this needed?

example/src/MutableReference/Prop.hs
```diff
@@ -3,6 +3,7 @@
 {-# LANGUAGE FlexibleInstances  #-}
 {-# LANGUAGE GADTs              #-}
 {-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE TypeOperators      #-}
```
stevana: Same here?

example/src/MutableReference/Prop.hs
```diff
@@ -169,23 +176,29 @@ prop_shrinkParallelMinimal = shrinkPropertyHelper (prop_referencesParallel RaceC
       , Internal (Inc   ref0)   (Symbolic var1)
       ]
 
-instance Read (Reference Symbolic (Opaque (IORef Int))) where
-  readPrec     = Reference . Symbolic <$> readPrec
-  readListPrec = readListPrecDefault
-
-instance Read (Internal Action) where
+instance Read (Untyped Action) where
```
stevana: Can we remove this read instance completely now?

example/src/MutableReference/Prop.hs
```diff
@@ -169,23 +176,29 @@ prop_shrinkParallelMinimal = shrinkPropertyHelper (prop_referencesParallel RaceC
...
+      appPrec = 10
 
-  readListPrec = readListPrecDefault
 
 deriving instance Eq1 v => Eq (Action v resp)
-
-instance Eq (Internal Action) where
-  Internal act1 sym1 == Internal act2 sym2 =
-    cast act1 == Just act2 && cast sym1 == Just sym2
+instance Eq (Untyped Action) where
```
stevana: I'm a bit confused here. Why do we need equality for untyped actions? I thought we needed equality of programs (which in turn consist of internal actions) for checking if counterexamples are minimal?

example/src/TicketDispenser.hs
```diff
@@ -226,17 +231,18 @@ prop_ticketDispenserParallelBad =
 
 -- Instances needed for the last property.
 
-instance Read (Internal Action) where
-
+instance Read (Untyped Action) where
```
stevana: Should no longer be needed?

Lysxia:
Ah, oops, I mistyped a `git add` and forgot to reset it.

example/src/MutableReference/Prop.hs
Lysxia: Right, read instances are no longer needed (except if users somehow need to read programs).

example/src/MutableReference/Prop.hs
Lysxia: Equality of `Internal Action` now defers to equality of `Untyped Action` to compare the first field (which I realize could just use a cast here). The pattern I'm going for here is that the user only needs to derive instances for either `Action` or `Untyped Action`.

example/src/DieHard.hs
Lysxia: No longer.

example/src/MutableReference/Prop.hs
stevana: Ah that makes sense. Deriving equality for `Action` and then defining it for `Untyped Action` using cast would be nice. (And then we could template Haskell away that later.)

example/src/MutableReference/Prop.hs
stevana: Let's make it easy for ourselves and assume they don't need to read programs, :-).

stevana:
(We might need to tweak the Travis timeout and/or run less of some slow tests...)

PR #161, opened 2017-10-02
Template haskell for HFunctor, HFoldable, HTraversable, Constructors, and shrinkers

Lysxia:


src/Test/StateMachine/Utils.hs
```diff
@@ -0,0 +1,13 @@
+module Test.StateMachine.Utils where
```
stevana: There's already a module called `Test.StateMachine.Internal.Utils`, perhaps it would make sense to put those things there?

stevana:
Also, would it make sense to also derive:
```haskell
instance Show (Untyped Action) where
   show (Untyped act) = show act
```
?

stevana:
(Also it would be good to add a changelog entry about this. And make a new release soon.)

src/Test/StateMachine/Utils.hs
Lysxia: Yeah that makes sense. I was trying to make a distinction between these utils for internal use, and the others as a convenience for users. But some of these functions might be useful in both categories, so it may be better instead to put everything in one module used internally and have `Test.StateMachine.Utils` reexport a relevant subset of it.

Lysxia:
Ah yes, I forgot to try `--pedantic`.  
I wasn't too sure about `Show`, because Haskell can already derive it, but now that I look at it again it is still a bit of boilerplate specific to this library that we can get rid of. I'll add the template haskell.

    deriving instance Show1 v => Show (Action v)
    deriving instance Show (Untyped Action)

And let me add the changelog entry! I'd also like to write some documentation about the new features this week-end, so I would be ready for a release next week.

Lysxia:
### TODO

- [ ] Show instances
- [ ] Changelog
- [ ] Headers for new files

src/Test/StateMachine/TH.hs
```diff
@@ -1,15 +1,49 @@
+-----------------------------------------------------------------------------
+-- |
+-- Module      :  Test.StateMachine.TH
+-- Copyright   :  (C) 2017, ATS Advanced Telematic Systems GmbH
```
stevana: I think you should be on the copyright for these files, as they contain your work! (Also feel free to add yourself to other modules where you feel you've contributed.)

Lysxia: Should I just add my name after ATS then?

Lysxia:
Hurray! I'm done for now. Feel free to prepare the next release!

jystic:
🚀 very nice, might need to steal some of this stuff for hedgehog

Lysxia:
Sure! Perhaps this could be split into a few reusable libraries?

stevana:
I think it would be nice to have a separate package for hfunctor and friends.

(Could be worth thinking about other things that can be shared between the implementations, e.g. histories and linearisation.)

jystic:
Yeah great idea!

PR #163, opened 2017-10-06
Feat/failing semantics

stevana:


src/Test/StateMachine/Types.hs
```diff
@@ -105,8 +109,13 @@ type Postcondition model act = forall resp.
 --   check.
 type InitialModel m = forall (v :: * -> *). m v
 
+data Result resp err = Ok resp | Fail err
```
Danten: Maybe have the arguments in opposite order? In case we want `Result` to be a a monad or whatever in the future?

stevana: Good point! Will fix in next PR.

PR #171, opened 2017-10-09
First go at Z inspried definition of relations and associcated operations.

stevana:
The idea is to use this little language to define concise and showable models.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,10 @@
-#### ???
+#### 0.2.0
+
+  * Z-inspired definition of relations and associated operations were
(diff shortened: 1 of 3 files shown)


stevana:
I'd like to get this in before the next release, so I can experiment with it. If it makes sense I'll move it to its own package. Also, I've created a separate issue for discussion (#172).

Lysxia:
The examples could be split into a separate file. Apart from that, I can't really comment about the high-level idea.

stevana:
I made doctests out of the small examples. I'll add a bigger example show casing the module later (when I figured out how to use it)...

PR #173, opened 2017-10-11
Refactor/hlint

stevana:


src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+singleton x y = [(x, y)]
+
+domain :: Rel a b -> [a]
+domain xys = [ x | (x, _) <- xys ]
+
+codomain :: Rel a b -> [b]
+codomain xys = [ y | (_, y) <- xys ]
+
+isTotalRel :: Eq a => Rel a b -> [a] -> Bool
+isTotalRel r xs = domain r == xs
```
Danten: Can't use `==` since for example
```haskell
isTotalRel (identity [0,1]) [0] == False
```
Which seems wrong to me, I think the spec should be that `xs` is a subset of the `domain r`

src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+
+isSurjRel :: Eq b => Rel a b -> [b] -> Bool
+isSurjRel r ys = codomain r == ys
```
Danten: Same here

src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+acs <||> bds =
+  [ ((a, b), (c, d))
+  | (a, c) <- acs
+  , (b, d) <- bds
+  ]
+
+------------------------------------------------------------------------
+
+isPartialFun :: (Eq a, Eq b) => Rel a b -> Bool
+isPartialFun f  = (f `compose` inverse f) == identity (codomain f)
```
Danten: This `==` seems wrong to me

```haskell
isPartialFun [(0,0), (0,0)] == False
```

src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+
+isTotalFun :: (Eq a, Eq b) => Rel a b -> [a] -> Bool
+isTotalFun r xs = isPartialFun r && domain r == xs
```
Danten: Instead of
```haskell
domain r == xs
```
use
```haskell
isTotalRel r xs
```

src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+
+isPartialInj :: (Eq a, Eq b) => Rel a b -> Bool
+isPartialInj r = isPartialFun r && isPartialFun (inverse r)
+
+isTotalInj :: (Eq a, Eq b) => Rel a b -> [a] -> Bool
+isTotalInj r xs = isTotalFun r xs && isPartialFun (inverse r)
+
+isPartialSurj :: (Eq a, Eq b) => Rel a b -> [b] -> Bool
+isPartialSurj r ys = isPartialFun r && codomain r == ys
```
Danten: Use
```haskell
isSurjRel r ys
```

src/Test/StateMachine/Z.hs
```diff
@@ -0,0 +1,189 @@
...
+-- |
+-- Module      :  Test.StateMachine.Z
+-- Copyright   :  (C) 2017, ATS Advanced Telematic Systems GmbH
+-- License     :  BSD-style (see the file LICENSE)
+--
+-- Maintainer  :  Stevan Andjelkovic <stevan@advancedtelematic.com>
+-- Stability   :  provisional
+-- Portability :  portable
+--
+-- This module contains Z-inspried combinators for working with relations. The
```
Danten: Why not just call it relations? Why mention `Z` at all?
https://en.wikipedia.org/wiki/Relation_(history_of_concept)#Logical_relations_and_database_theory
People have used relations before `Z`...

src/Test/StateMachine/Z.hs
stevana: The idea is that since we can't enumerate all of `a` (we would need a `Bounded a` and `Enum a` constraint to do that), so we ask the user to supply the universe, i.e. `xs`, against which we check totality.

Arguebly order shouldn't matter though, e.g. `isTotalRel (identity [0, 1]) [1, 0]` should be true... I guess I'll introduce a newtype wrapper for `Rel` with a custom equality or perhaps use `Set`... I like the ergonomics of lists though...

src/Test/StateMachine/Z.hs
stevana: Many operations are taken straight from Z, e.g. overriding and subtraction, which are not part of standard relational terminology.

src/Test/StateMachine/Z.hs
Danten: For the kind of relations you are modelling here order does not matter, I don't really think there is an argument otherwise.

Or just use a different function instead of `==` here
```haskell
relEq :: Eq a => [a] -> [a] -> Bool
xs `relEq` ys = xs `isSubsetOf` ys && ys `isSubsetOf` xs
```

stevana: Yeah that's a good idea, thanks.

src/Test/StateMachine/Z.hs
stevana: Fixed, thanks!

src/Test/StateMachine/Z.hs
stevana: Fixed, thanks!

PR #180, opened 2017-10-13
Let the post-condition have access to failed results.

stevana:


src/Test/StateMachine.hs
```diff
@@ -279,7 +279,7 @@ runParallelProgram'
 runParallelProgram' n StateMachine {..} prog =
   replicateM n $ do
     hist <- run (executeParallelProgram semantics' prog)
-    return (hist, linearise transition' postcondition' model' hist)
+    return (hist, linearise transition' (okPostcondition postcondition') model' hist)
```
Danten: Oh, isn't this screwing up `linearise`? It doesn't seem to be that difficult to change it to do the proper things for `postcondition`s that can react to `Fail`.

stevana: Fixed, thanks!

PR #182, opened 2017-10-23
Example/bank2

stevana:
Start work on refinement example, tweak output and how failing semantics is handled.

Sorry about this big messy PR, I started working on this bank/refinement example and ended up doing a bunch of other stuff at the same time.

* Output is now closer to what is described in issue #77, if `prettyProgram'` is used;

* `executeProgram'` returns a `Result`, like Erlang qc... I believe this is needed to handle failing semantics properly;
 
* `StateMachine''` uses boolean post-conditions and a modified transition function that can handle failing semantics (the idea is that this will replace the current `StateMachine'`).

It would be good to hear your thoughts, before I refactor the other examples to use this new interface!


src/Test/StateMachine/Types.hs
```diff
@@ -95,6 +102,24 @@ data StateMachine' model act err m = StateMachine
...
+data StateMachine'' model act err m = StateMachine''
+  { generator''     :: Generator model act
+  , shrinker''      :: Shrinker  act
+  , precondition''  :: Precondition model act
+  , transition''    :: Transition' model act err
+  , postcondition'' :: Postcondition'' model act err
+  , model''         :: InitialModel model
+  , semantics''     :: Semantics' act err m
+  , runner''        :: Runner m
+  }
```
Lysxia: Is there a way to merge `StateMachine'` and `StateMachine''`?

Lysxia: Ah, nevermind, I didn't see you say

> the idea is that this will replace the current StateMachine'

PR #183, opened 2017-10-24
First go at creating parallel program from sequential one.

stevana:
Work in progress; PR created for feedback.


stevana:
The code here is based Hans' answer about how Erlang QuickCheck does it and on: "The basic idea is to define the assertion {P} S {Q} to mean that if execution is begun anywhere in S with P true, then P will remain true until S terminates, and Q will be true if and when S terminates." -- Lamport, https://link.springer.com/article/10.1007/BF00289062 .

PR #186, opened 2017-11-27
Add Show1 constraint to transition.

stevana:


Lysxia:
Why is it necessary?

stevana:
I have a model that uses `Fun` from `Test.StateMachine.Z`, and in the transition function I'd like to use:

```haskell
(!) :: (Eq a, Show a, Show b) => Fun a b -> a -> b
f ! x = maybe (error msg) Prelude.id (lookup x f)
  where
  msg = "!: failed to lookup `" ++ show x ++ "' in `" ++ show f ++ "'"
```

So that I can see why the function application failed.

PR #189, opened 2017-12-06
First go at creating parallel program from sequential one.

stevana:
Big PR that tries to generate parallel programs from sequential ones (inspired by Hans' comment about how Erlang does it). The PR still needs some work, in particular the properties that assert that the minimal counterexamples are found are broken, but some feedback would be appreciated.


src/Test/StateMachine/Internal/Types.hs
```diff
@@ -79,6 +80,12 @@ deriving instance Read (Untyped act) => Read (ParallelProgram act)
 data Fork a = Fork a a a
   deriving (Eq, Functor, Show, Ord, Read)
 
+data ParallelProgram' act = ParallelProgram' (Program act) [Program act]
```
Lysxia: Is this going to replace the `Fork`-based `ParallelProgram`?

stevana: Yeah, that's the idea.

src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -239,3 +359,93 @@ toBoxDrawings prog = toBoxDrawings' allVars
...
+parallelSafe
+  :: Precondition model act
+  -> Transition' model act err
+  -> model Symbolic
+  -> Program act
+  -> Bool
+parallelSafe precondition transition model0
+  = and
+  . map (preconditionsHold model0)
+  . permutations
```
Lysxia: This doesn't seem right. Doesn't each thread run in program order?

Lysxia: My understanding is that `parallelSafe` is testing just one thread at a time.

Lysxia:
Perhaps a good property to want of generated programs is that any interleaving of a parallel program produces a valid sequential program?

src/Test/StateMachine/Internal/Parallel.hs
stevana: Something that's perhaps a bit confusing is that the list of programs (suffixes) in: `ParallelProgram' act = ParallelProgram' (Program act) [Program act]` isn't of length 2 (for two threads)... The idea is that every program (or list of internal actions) in that list can be run in parallel. Perhaps the type should be `ParallelProgram' act = ParallelProgram' (Program act) [[Internal act]]` and a comment added that says that the inner list is parallel safe to make things more clear...

stevana: By the way, I thought about how to generate only two suffixes (one per worker), but I couldn't come up with a good way (because I agree that it would be more intuitive that way). Ideas?

Lysxia: Each suffix is a program for one worker which runs it in sequence, and workers can be run in parallel. But here it seems we are permuting the actions of one worker.

stevana: My understanding is that each program in the suffix list gets split up between the workers upon execution (again I agree that this isn't intuitive given the type), and therefore this check makes sense!? :-)

stevana:
I agree, this would be a good property to have. I'll try to express it in code!

src/Test/StateMachine/Internal/Parallel.hs
Lysxia: Ah! I got unstuck from thinking the `[ParallelProgram']` is the transpose of what it really is. Now it makes sense...

stevana: Good :-). 

@Danten suggested moving the splitting from the point of execution to the type itself, something like: `ParallelProgram act = ParallelProgram (Program act) [(Program act, Program act)]` to make things more clear...

src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -94,29 +94,42 @@ generateParallelProgram generator precondition transition model = do
...
+  shrinkSuffixes :: [(Program act, Program act)] -> [[(Program act, Program act)]]
+  shrinkSuffixes = shrinkList (shrinkPair shrinkProgram')
+
+  shrinkMoveSuffixToPrefix :: [ParallelProgram act]
+  shrinkMoveSuffixToPrefix = case suffixes of
+    []                   -> []
+    (suffix : suffixes') ->
+      [ ParallelProgram (prefix <> Program [prefix'])
+                        (bimap Program Program
+                              (splitAt (length suffix' `div` 2) suffix') : suffixes')
```
Danten: Hmm this seems wrong to me, can't this potentially move things from one parallel part to the other?

src/Test/StateMachine/Internal/Parallel.hs
```diff
@@ -94,29 +94,42 @@ generateParallelProgram generator precondition transition model = do
...
+      | (prefix', suffix') <- pickOneReturnRest (unProgram (fst suffix) ++
+                                                 unProgram (snd suffix))
+      ]
+
```
Danten: something like
```
pickOneReturnRestTupleEdition :: ([a], [a]) -> [(a, ([a],[a]))]
pickOneReturnRestTupleEdition (xs, ys) = 
  map (id *** flip (,) ys) (pickOneReturnRest xs)
  map (id *** (,) xs) (pickOneReturnRest ys)
```

src/Test/StateMachine/Internal/Parallel.hs
stevana: @Lysxia: Changed the type around, do you find it easier to read now?

Lysxia: Yes indeed :+1:

PR #247, opened 2018-12-06
tame -Wstar-is-type and prepare for ghc 8.6

kderme:
Since ghc 8.6.2 the use of `*` is considered deprecated and raises `-Wstar-is-type`. At this pr we follow the recommended fix (more https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/using-warnings.html#ghc-flag--Wstar-is-type).


stevana:
Thanks for the fix!

PR #252, opened 2018-12-18
Replace MonadBaseControl IO with MonadUnliftIO

mrBliss:
* Replace `MonadBaseControl IO m` with `MonadUnliftIO m`. Reasoning: see the  documentation for [`unliftio`](http://hackage.haskell.org/package/unliftio).

* Replace `liftBaseWith (const ..)` with `liftIO ..`. This means that these  `MonadBaseControl IO m` constraints can be simplified to `MonadIO m`  constraints.

* Remove manual `liftIO`s where possible by using lifted functions from  `unliftio`.

* Replace the uses of lifted functions from `lifted-async` by the  corresponding lifted variants from `unliftio`.

* Replace the uses of functions from `stm` by the corresponding lifted  variants from `unliftio`.

### DIFF
--- a/quickcheck-state-machine.cabal
+++ b/quickcheck-state-machine.cabal
@@ -50,16 +50,14 @@ library
         base >=4.9 && <5,
         containers >=0.5.7.1,
         exceptions >=0.8.3,
-        lifted-async >=0.9.3,
         matrix >=0.3.5.0,
-        monad-control >=1.0.2.2,
         mtl >=2.2.1,
         pretty-show,
         QuickCheck >=2.9.2,
         split >=0.2.3.3,
-        stm >=2.4.4.1,
         tree-diff >=0.0.2,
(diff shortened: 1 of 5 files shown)


mrBliss:
@stevana Thanks for the quick merge!

However, I was expecting some resistance, to be honest :slightly_smiling_face:, as this change might break things for existing users. More concretely, users that use a custom monad `m` for their `StateMachine` previously had to provide an instance of `MonadBaseControl IO m`, but after this change they will have to provide an instance of `MonadUnliftIO m` instead. What is your opinion on this?

stevana:
I think this library still doesn't have that many users (especially affected by this change), and so I'd rather break and improve things than be stuck with bad decisions. (If anyone who's reading this is affected, please let us know).

We should add this to the change log and make a major version bump next release though.

PR #253, opened 2019-01-02
Add a stopcondition to StateMachine

mrBliss:
This stopcondition stops the test in progress. Imagine that an (expected) error is thrown during the tests and the tested system is now in a state where it can no longer process new commands. In that case we should be able to gracefully stop the test in progress.


mrBliss:
@stevana A backwards-compatible alternative suggested by @edsko is to simply stop when no pre-condition holds, instead of reporting  a deadlock. Labelling can be used to detect unexpected cases where no pre-condition holds.

I'll open a PR for this alternative too.

PR #242, opened 2018-11-05
Feat/markov chain

stevana:

Issue #236, opened 2018-10-22
What should mock be used for


akegalj:
I am not sure what should be `mock` , last parammeter to `StateMachine` be used for and how it is used in a current implementation?

This line is onpar with my understanding https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/test/Echo.hs#L146

Also I can't deduce why `mock` would be needed from readme section on hackage:
```
Sequential property
The sequential property checks if the model is consistent with respect to the semantics. The way this is done is:

generate a list of actions;

starting from the initial model, for each action do the the following:

check that the pre-condition holds;
if so, execute the action using the semantics;
check if the the post-condition holds;
advance the model using the transition function.
If something goes wrong, shrink the initial list of actions and present a minimal counterexample.

Parallel property
The parallel property checks if parallel execution of the semantics can be explained in terms of the sequential model. This is useful for trying to find race conditions -- which normally can be tricky to test for. It works as follows:

generate a list of actions that will act as a sequential prefix for the parallel program (think of this as an initialisation bit that setups up some state);

generate two lists of actions that will act as parallel suffixes;

execute the prefix sequentially;

execute the suffixes in parallel and gather the a trace (or history) of invocations and responses of each action;

try to find a possible sequential interleaving of action invocations and responses that respects the post-conditions.

The last step basically tries to find a linearisation of calls that could have happend on a single thread.
```

stevana:
The way `mock` should be used is that it should return a mocked response to each
command using values from the model, e.g. `Read` from the README looks up the
value associated to the reference we are trying to read from in the model. If a
command returns `Reference`s, e.g. `Create` from the README, then `genSym`
should be used to create those (`Symbolic`) `Reference`s.

Why `mock` is needed is more complicated.

Most importantly, is that it solves the problem of allowing more than one
`Reference` in a response. It's a hack however and I wish I could find a better
solution.

Another reason for not removing it is, because I've also played around with the
idea of using `mock` to create a mocked APIs. I've still not worked out the
details of how exactly this should work though. (Because this isn't actually
used by the library yet, it's not important for `mock` to return the correct
responses, e.g. having `mock` for `Read` return `ReadValue 0` is completely
fine.)

Does that make sense?

I'll update the README to include more info about `mock` while I keep thinking
of how to get rid of it or actually use it for something meaningful...

stevana:
[...]

PR #254, opened 2019-01-02
Gracefully stop when no precondition holds

mrBliss:
Imagine that an (expected) error is thrown during the tests and the tested system is now in a state where it can no longer process new commands. In that case we should be able to gracefully stop the test in progress.

Simply stop the test when no precondition holds instead of reporting a deadlock error. Labelling can be used to detect such deadlocks.

This is an alternative to #253.

### DIFF
--- a/quickcheck-state-machine.cabal
+++ b/quickcheck-state-machine.cabal
@@ -104,6 +104,7 @@ test-suite quickcheck-state-machine-test
                        CrudWebserverDb,
                        DieHard,
                        Echo,
+                       ErrorEncountered,
                        MemoryReference,
                        TicketDispenser
 
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -79,7 +79,7 @@ import           Test.StateMachine.Utils
 ------------------------------------------------------------------------
 
 forAllParallelCommands :: Testable prop
-                       => (Show (cmd Symbolic), Show (model Symbolic))
+                       => (Show (cmd Symbolic))
                        => (Generic1 cmd, GConName1 (Rep1 cmd))
                        => (Rank2.Foldable cmd, Rank2.Foldable resp)
                        => StateMachine model cmd m resp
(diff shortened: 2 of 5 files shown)


mrBliss:
The non-determinism is not an issue (anymore) for me, I just want to be able to stop a test.

So:
```haskell
model Symbolic -> Maybe (Gen (cmd Symbolic))
```
is a perfectly good alternative for me.

Do you want me to make a PR for this or do you want to wait for #242?

stevana:
How soon do you need this?

Merging #242 might take some time. I don't exactly know how to provide the user with the choice of using the old size based generation or the Markov chain based one. Perhaps by splitting the `StateMachine` record into two, one simple interface (size based and no `mock`* nor perhaps `invariant`) and an advanced/experimental interface (Markov chain based and with `mock` and `invariant`). The idea would be that the simple interface is enough for the examples in `test/` and that it's relatively stable, while the advanced/experimental is more of a playground for new ideas where the interface might change more often. But I don't see how to do this without a lot of code duplication though (but perhaps that's not such a big deal).

If you can wait or use the #242 branch, then I'd prefer to not change the type of `generator`. If not, then I'd also be happy to merge the `Maybe (Gen (cmd Symbolic)` change, but maybe I'd revert it once #242 gets merged to keep the simple interface simple.

Thoughts?

*: `mock` is confusing, https://github.com/advancedtelematic/quickcheck-state-machine/issues/236, and not necessary as can be witnessed by old [versions](https://github.com/advancedtelematic/quickcheck-state-machine/blob/v0.1.0/src/Test/StateMachine/Types.hs) of this library (unless you need responses to be able to return multiple references).

mrBliss:
I think we need this before the Markov-chain stuff is finished.

I'll send a PR for the change (that you may revert later once #242 is merged).

It would be great if you made a release that includes both the PR that I'll open and Edsko's #255.

edsko:
This is neither here nor there, but..

> *: mock is confusing, #236 

Actually, it's not that confusing, as I hope to show in my blog post :)

stevana:
> It would be great if you made a release that includes both the PR that I'll open and Edsko's #255.

No problem.

> Actually, it's not that confusing, as I hope to show in my blog post :)

Looking forward to reading it!

Issue #262, opened 2019-01-07
Make model available to shrinker


edsko:
At the moment, the type of the shrinker is 

```haskell
shrinker :: cmd Symbolic -> [cmd Symbolic]
```

This means the shrinker for commands has no context available at all. When we take shrinking serious, this is quite a limitation. I've been trying to workaround this in a million different ways but I get stuck each time. 

I propose that we generalize the type to

```haskell
shrinker :: model Symbolic -> cmd Symbolic -> [cmd Symbolic]
```

so that the shrinker has the context available. Initially I thought that this would make shrinking more expensive, which is why I didn't go down this route before. _But_ thinking about it more I now realize that this is not actually true: after all, shrinking is _always_ followed by a validation step, so we compute the model _anyway_ for all shrinks. 

What this means is that shrinking cannot be the current `shrinkList`-then-`validCommands`; instead it must combine the two. 

@stevana I will start working on a patch to do this. If you think this is a terrible idea, let me know ASAP, but I think it will make the library more general, without making shrinking more expensive.

stevana:
I think this generalisation makes sense.

edsko:
Awesome! Then I will work on a patch :)

stevana:
Fixed in referenced PR.

PR #263, opened 2019-01-07
WIP: Generalize shrinking

edsko:
See #262 for discussion.

@stevana This is a WIP PR. This implements the generalization we previously discussed, but only for the sequential case. Submitting it as a PR now so you can take a look to see what you think. I've not yet tried the test suite (since that will require implementing the parallel case also), but I _have_ tested it in my real project and it works like a charm. As you can see in the PR, the update to the code is actually not that big either, which is good. 

I think tomorrow I will try to finish my own test case now that this is working and available before trying to update the parallel case. I've not used parallel testing very much yet so might be a little slower.

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -22,7 +22,7 @@ module Test.StateMachine.Parallel
   ( forAllParallelCommands
   , generateParallelCommands
   , shrinkParallelCommands
-  , validParallelCommands
+  , shrinkAndValidateParallel
   , prop_splitCombine
   , runParallelCommands
   , runParallelCommandsNTimes
@@ -39,16 +39,15 @@ import           Control.Monad
 import           Control.Monad.Catch
                    (MonadCatch)
 import           Control.Monad.State
-                   (State, evalState, put, runStateT)
+                   (runStateT)
 import           Data.Bifunctor
                    (bimap)
 import           Data.List
                    (partition, permutations)
 import           Data.List.Split
-                   (splitPlacesBlanks)
-import           Data.Map
-                   (Map)
-import qualified Data.Map                          as M
+                   (splitPlaces, splitPlacesBlanks)
+import           Data.Maybe
+                   (mapMaybe)
 import           Data.Monoid
                    ((<>))
 import           Data.Set
@@ -58,8 +57,7 @@ import           Data.Tree
                    (Tree(Node))
 import           Prelude
 import           Test.QuickCheck
-                   (Gen, Property, Testable, choose, property,
-                   shrinkList, sized)
+                   (Gen, Property, Testable, choose, property, sized)
 import           Test.QuickCheck.Monadic
                    (PropertyM, run)
 import           Text.PrettyPrint.ANSI.Leijen
@@ -166,25 +164,30 @@ shrinkParallelCommands
   => Rank2.Foldable resp
   => StateMachine model cmd m resp
   -> (ParallelCommands cmd -> [ParallelCommands cmd])
-shrinkParallelCommands sm@StateMachine { shrinker, initModel }
+shrinkParallelCommands sm@StateMachine { initModel }
                        (ParallelCommands prefix suffixes)
-  = filterMaybe (flip evalState (initModel, M.empty, newCounter) . validParallelCommands sm)
-      [ ParallelCommands prefix' (map toPair suffixes')
-      | (prefix', suffixes') <- shrinkPair' shrinkCommands' shrinkSuffixes
-                                            (prefix, map fromPair suffixes)
+  = concatMap go
+      [ Shrunk s (ParallelCommands prefix' (map toPair suffixes'))
+      | Shrunk s (prefix', suffixes') <- shrinkPairS shrinkCommands' shrinkSuffixes
+                                                     (prefix, map fromPair suffixes)
       ]
       ++
       shrinkMoveSuffixToPrefix
   where
-    shrinkCommands' :: Commands cmd -> [Commands cmd]
-    shrinkCommands'
-      = map Commands
-      . shrinkList (liftShrinkCommand shrinker)
-      . unCommands
+    go :: Shrunk (ParallelCommands cmd) -> [ParallelCommands cmd]
+    go (Shrunk shrunk cmds) =
+        shrinkAndValidateParallel sm
+                                  (if shrunk then DontShrink else MustShrink)
+                                  (initValidateEnv initModel)
+                                  cmds
 
-    shrinkSuffixes :: [(Commands cmd, Commands cmd)] -> [[(Commands cmd, Commands cmd)]]
-    shrinkSuffixes = shrinkList (shrinkPair shrinkCommands')
+    shrinkCommands' :: Commands cmd -> [Shrunk (Commands cmd)]
+    shrinkCommands' = map (fmap Commands) . shrinkListS' . unCommands
 
+    shrinkSuffixes :: [(Commands cmd, Commands cmd)] -> [Shrunk [(Commands cmd, Commands cmd)]]
+    shrinkSuffixes = shrinkListS (shrinkPairS' shrinkCommands')
+
+    -- Moving a command from a suffix to the prefix preserves validity
     shrinkMoveSuffixToPrefix :: [ParallelCommands cmd]
     shrinkMoveSuffixToPrefix = case suffixes of
       []                   -> []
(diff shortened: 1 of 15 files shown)


stevana:
Yep, looks good!

(If you can think of a way to include some tests of shrinking working along the way that would be really useful. The bug you found before was a regression actually... We used to have some tests for shrinking, but those tests were annoying to maintain so I deleted them during a refactor...)

Regarding parallel testing: that's actually the cool part, :-). Feel free to ask, if anything isn't clear!

edsko:
@stevana Ok, just pushed another commit that deals with the parallel case. Please take a careful look. I think the modifications in the end are similar as for the sequential case, and I think apart from the shrinking itself, it should not change any of the current behaviour. The test suite passes except that for the doc test I'm getting

```
# quickcheck-state-machine-test -p Doctest
Tests
  Doctest Z module: FAIL
    Exception: quickcheck-state-machine-test: can't find a package database at /home/edsko/wt/clients/iohk/ouroboros-network/dist-newstyle/packagedb/ghc-8.2.2
```

which seems completely unrelated to my patch? 

A comment on the implementation: the sequential case split the shrinking of a list of commands from the 
shrinking of the commands themselves (which needs the model). For the parallel case we need to do the same, except that we now have more structure: a prefix, then a list of pairs of list of commands. To keep this somewhat tractible, I introduced some machinery in the `.Util` module to shrink a value and record whether something changed. This then makes the rest of the implementation fairly straight-forward.

Don't merge this yet, I will try to also write a property or two about shrinking. However, I'd be curious to hear your review of my changes and whether you think this is OK.

edsko:
I may have messed something up... the shrinking test that I'm writing is stress testing references, and it's failing somewhere. Don't know yet if my test is broken or the code. Will push the test as part of this commit also, of course.

src/Test/StateMachine/Sequential.hs
```diff
@@ -196,58 +200,56 @@ getUsedVars = Rank2.foldMap (\(Symbolic v) -> [v])
...
-validCommands :: forall model cmd m resp. (Rank2.Traversable cmd, Rank2.Foldable resp)
-              => StateMachine model cmd m resp -> Commands cmd
-              -> State (model Symbolic, Map Var Var, Counter) (Maybe (Commands cmd))
-validCommands StateMachine { precondition, transition, mock } =
-  fmap (fmap Commands) . go . unCommands
+-- | Shrink individual commands using specified shrinker
+--
+-- If we shrunk something already, this will return either the empty list
+-- (if validation fails) or a singleton list (otherwise). If not, it will
+-- shrink exactly one command in the list to @n@ shrink commands, and then
```
stevana: I don't understand this comment, is there a word missing perhaps?

stevana: Yes, thanks!

edsko:
@stevana Ok, I see my mistake. I don't call `shrinkAndValidate` on the prefix. Will fix. Have quite a nice property now that shrinking must satisfy, I think you'll like it :) Fix coming.

edsko:
Uh, no, sorry, that comment is wrong, and makes no sense. But shrinking _is_ going wrong and I do have a simple counterexample. I'll keep digging, sorry for misleading.

stevana:
Only had a quick look so far, but it looks neat! :-)

By the way, one more property of shrinking that might be worth testing is that: any result of shrinking should be possible to generate. (This is kind of what Hedgehog-style shrinking gets for free).

edsko:
> By the way, one more property of shrinking that might be worth testing is that: any result of shrinking should be possible to generate. (This is kind of what Hedgehog-style shrinking gets for free).

This is not true in general for `quickcheck-state-machine`. It is entirely possible to write combinations of generators and shrinkers such that the shrinker can shrink to something that the generator will never generate. In fact, my "real" project in which I'm applying all this stuff is an example of that.

What I _was_ thinking though is that it would be good to test that the model that the shrinker is given is the same one that you'd get if you then walk over the shrunk test case step by step, updating the model as you go. I think I have a way to do this but I need to ponder it a bit.

stevana:
> This is not true in general for quickcheck-state-machine. It is entirely possible to write combinations of generators and shrinkers such that the shrinker can shrink to something that the generator will never generate. In fact, my "real" project in which I'm applying all this stuff is an example of that.

Possible yes, but is it ever desirable? Intuitively this feels like a mismatch between generators and pre-conditions to me...

> What I was thinking though is that it would be good to test that the model that the shrinker is given is the same one that you'd get if you then walk over the shrunk test case step by step, updating the model as you go. I think I have a way to do this but I need to ponder it a bit.

Yeah, would be nice to test as well.

edsko:
> Possible yes, but is it ever desirable? Intuitively this feels like a mismatch between generators and pre-conditions to me...

Well, you can let me know what you think after I write my blog post :) 

But even for the simple model that I'm now using to test shrinking, it's not true: commands are given identifiers, which are generated as increasing integers. But after shrinking those integers won't be consecutive anymore. 

So at the very least it depends on the model, so we could only test this for a given model. Having said that, for the model that I'm working with to test shrinking, we could test this "up to command IDs", but I think it would be prohibitively expensive: you'd have to say, "given this shrunk set of commands, find a seed such that..". I don't think it would be feasible? :thinking:

edsko:
It's an interesting comment though. I feel it's actually an _advantage_ that we can make this distinction (that this property is not true), and if that is not the case in hedgehog, maybe this is one plus point for the QC approach. But I need to think more about it :)

stevana:
> Well, you can let me know what you think after I write my blog post :)

Yeah :-)

> So at the very least it depends on the model, so we could only test this for a given model. 

Perhaps this could be a meta-property that the library provides the user to sanity check their generators and pre-conditions?

>  it would be prohibitively expensive: you'd have to say, "given this shrunk set of commands, find a seed such that..". I don't think it would be feasible?

Not with the current way of generating, but it would be easier with the Markov chain approach:

  https://github.com/advancedtelematic/quickcheck-state-machine/pull/242/files#diff-cb1f5b1b1e7e0a159c0e9565f33ffb75R232

Start with the first command name, make sure it's reachable from the initial state of the chain, etc. Not sure if one can effectively check the arguments of the commands though...

Sorry, this got a bit off topic.

edsko:
It's an interesting discussion for sure! Perhaps we can consider this to be separate from this specific PR? I think I know how to fix shrinking, and I'll extend my test also to check the model, and then I think we might be in good shape. Do you think it warrants another release?

edsko:
@stevana Quick question for you: `advanceModel` returns a `(model Symbolic, Counter)` pair. Do we have the property that the `(model Symbolic, Map Var Var, Counter)` that we get from validation is the same model and counter?

edsko:
Oh, wait, my test is failing because it actually has an race condition, and `quickcheck-state-machine` found it! Amazing :grinning:

test/ShrinkingProps.hs
```diff
@@ -0,0 +1,519 @@
...
+    tests
+    -- Just to avoid compiler warnings
+  , ShrinkSeqFailure(..)
+  , ShrinkParFailure(..)
+  ) where
+
+import           Prelude                       hiding
+                   (elem)
+
+import           Control.Concurrent.STM
```
mrBliss: Maybe use `import UnliftIO` or `import UnliftIO.STM` here, see #253. You can also drop the dependency on `stm`.

stevana:
Alright, cool!

src/Test/StateMachine/Parallel.hs
```diff
@@ -87,6 +87,46 @@ forAllParallelCommands :: Testable prop
 forAllParallelCommands sm =
   forAllShrinkShow (generateParallelCommands sm) (shrinkParallelCommands sm) ppShow
 
+-- | Generate parallel commands.
```
mrBliss: @stevana Can you verify whether I got this right?

mrBliss:
I have spent some time trying to understand the generation of parallel commands. I think now that the bug is not there, but actually in the test, due to a misunderstanding of how these parallel commands are generated.

@stevana Maybe squash these commits before merging.

stevana:
> I think now that the bug is not there, but actually in the test [...]

Did you manage to fix the bug also? Is this PR ready for review now?

stevana:
Cool!

There still seem to be some simple problems with building as witnessed by Travis... I'll try to comment on the rest of the code during the day.

mrBliss:
Interesting, one CI build is failing because one of the new tests fails, which means there is a problem with parallel shrinking (or possibly with the test itself). I will investigate.

edsko:
Alright, I'm back :sunglasses: @mrBliss and I discussed what still needs fixing, and I'll try to get this to mergeable state. @stevana will ping you when it's ready :)

edsko:
@stevana Ok, tests are passing (I ran with `--quickcheck-tests 1000`, I'll run it in the background with 10000 tonight), and I think the code is correct now. Here's an overview of the final changes:

### `Command` now records the response

I've changed `Command` to

```haskell
-- | Previously symbolically executed command
--
-- Invariant: the variables must be the variables in the response.
data Command cmd resp = Command !(cmd Symbolic) !(resp Symbolic) ![Var]
```

Before we were only recording those variables, and not the response itself. However, every time we create a `Command`, we _have_ that response; we just throw it away. Keeping it has an important advantage: when we apply a previously generated `Command` to a model, we don't need the mock function anymore. This means that `advanceModel` changes to 

``` haskell
advanceModel :: StateMachine model cmd m resp
             -> model Symbolic      -- ^ The model.
             -> Commands cmd resp   -- ^ The commands.
             -> model Symbolic
advanceModel StateMachine { transition } model0 =
  go model0 . unCommands
  where
    go model []                              = model
    go model (Command cmd resp _vars : cmds) =
        go (transition model cmd resp) cmds
```

where we don't need the counter anymore. Not only does this make the code a little cleaner, it also removes some head-scratching: in the old `advanceModel`, we were _ignoring_ the `vars` associated with the command, creating a _new_ response, and then continuing with `cmds`... but what if those `vars` were different from the ones we generated before, and there are commands in `cmds` that referred to those old variables? _If_ we are careful that we apply this generation in precisely the same way we should be okay, but this is rather brittle. By recording the response this whole problem goes away.

### The validator for parallel commands interleaves left and right parts of the suffixes

The old validator was applying the sequential validator to 

* `prefix <> mconcat allLeftSuffixes`
* `prefix <> mconcat allRightSuffixes`

basically. But this is not correct. A parallel program such as

```
[ create reference 0 || create reference 1 ]
[ use reference 1    || ..                 ]
```

is perfectly correct: at the start of the second parallel suffix, we can assume _all_ commands (both left and right) have been executed. Initially I thought that this was not a valid program, which is why I thought above that the generator for parallel commands was broken, but that was a misunderstanding on my part. My parallel shrinker however made this same assumption as the old parallel shrinker did, which is the bug that @mrBliss identified. This also means that the old validator (before this PR) was throwing away too many programs: basically it required that left and right suffixes were completely independent.

The right thing to do is validate in this manner:

``` haskell
modelAfterPrefix <- validate initModel prefix

modelAfterLeft1  <- validate modelAfterPrefix left1
_                <- validate modelAfterPrefix right1
let modelAfterSuffix1 = advance modelAfterLeft1 right1

modelAfterLeft2  <- validate modelAfterPrefix left2
_                <- validate modelAfterPrefix right2
let modelAfterSuffix2 = advance modelAfterLeft2 right2

..
```

which is what this PR implements.

This has the additional benefit that we don't need to have the final `parallelSafeMany` check: since we are precise in the models that we use during parallel validation, this final check becomes superfluous.

(Bonus points: we don't need the dependency on the `split` package anymore :)

### Other consequences

@mrBliss already updated the shrinking properties to reflect this better understanding of how parallel commands work, but the "do we get the correct model" test also needed to be updated, which I've done.

Finally, the change to `Command` has some consequences for user-land code also; previously it was difficult to define

``` haskell
execCmd :: Model Symbolic -> QSM.Command Cmd Resp -> Model Symbolic
```

but now it's easy: since the command carries its response, this is just a call to `transition`.

edsko:
It took a little over 2 hours but 10,000 tests passed successfully.

edsko:
@stevana I think this is ready :)

src/Test/StateMachine/Sequential.hs
```diff
@@ -193,69 +196,126 @@ getUsedVars :: Rank2.Foldable f => f Symbolic -> [Var]
...
+          -- shrink at most one command
+          let candidates :: [(ShouldShrink, cmd Symbolic)]
+              candidates =
+                case shouldShrink of
+                  DontShrink -> [(DontShrink, remapped)]
+                  MustShrink -> map (DontShrink,) (shrinker model remapped)
+                             ++ [(MustShrink, remapped)]
+          in flip concatMap candidates $ \(shouldShrink', cmd) ->
+               if boolean (precondition model cmd)
+                 then let (resp, counter') = runGenSym (mock model cmd) counter
```
stevana: Can we get rid the use of `mock` here as well?

edsko: No, we cannot: since we are executing the command in a different model, it may return a different result, so we really need the `mock` function to get us a new response.

edsko:
Awesome! :+1: :grin:

PR #279, opened 2019-03-01
Add existsCommands to generate commands from a list of generators.

stevana:


balajirrao:
@stevana could you add an example, may be ? if it is easy enough..

PR #291, opened 2019-03-29
We should check for the mock-semantic mismatch before writing the

stevana:
response to the history.


balajirrao:
Looks good.

To make it more obviously correct, I would refactor the `go` to have the two `writeTChan`s at only once place to make it easy to see that exactly two things are written to the channel in all cases.

PR #293, opened 2019-04-02
Add Tower of Hanoi example

abonie:
Works for N discs and 3 pegs, but realisticly more than 4 discs will
cause it to struggle. There is nothing smart about this method.

### DIFF
--- a/test/Hanoi.hs
+++ b/test/Hanoi.hs
@@ -1,0 +1,127 @@
+{-# LANGUAGE KindSignatures     #-}
+{-# LANGUAGE DeriveAnyClass     #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE FlexibleInstances  #-}
+
+------------------------------------------------------------------------
+
+-- Solution to the famous Tower of Hanoi puzzle using tools for state
+-- machine property-based testing.
+--
+-- The puzzle is to move N discs of different sizes from one peg to
+-- another, with one auxiliary peg and a restriction that no disc may ever
+-- be placed on top of a smaller disc. Only one disc can be moved at a time.
+
+------------------------------------------------------------------------
+
+module Hanoi
+  ( Command(..)
+  , Model(..)
+  , initModel
+  , transitions
+  , prop_hanoi
+  ) where
+
+import Data.Kind
+         (Type)
+import Data.Array
+import Data.Maybe
+import Data.TreeDiff.Expr ()
+import GHC.Generics
+         (Generic, Generic1)
+import Prelude
+import Test.QuickCheck
+         (Gen, Property, (===), Arbitrary (arbitrary), choose, suchThat)
+import Test.QuickCheck.Monadic
+         (monadicIO)
+import Test.StateMachine
+import qualified Test.StateMachine.Types.Rank2 as Rank2
+
+------------------------------------------------------------------------
+
+-- The model keeps track of which disc is on which peg
+
+newtype Model (r :: Type -> Type) = Model (Array Int [Int])
+  deriving (Show, Eq, Generic)
+
+-- There are 3 pegs, so the bounds are (0, 2)
+
+pegsBounds :: (Int,Int)
+pegsBounds = (0, 2)
+
+instance ToExpr (Model r) where
+  toExpr (Model a) = toExpr $ elems a
+
+initModel :: Int -> Model r
+initModel discs = Model $ listArray pegsBounds [[1..discs], [], []]
+
+-- Allowed action is to move one disc from the top of one peg to the top of another
+
+data Command (r :: Type -> Type) = Move (Int,Int)
+  deriving (Eq, Show, Generic1, Rank2.Functor, Rank2.Foldable, Rank2.Traversable, CommandNames)
+
+instance Arbitrary (Command r) where
+  arbitrary = do
+    x <- choose pegsBounds
+    y <- choose pegsBounds `suchThat` (/= x)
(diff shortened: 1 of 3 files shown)


test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
+{-# LANGUAGE KindSignatures     #-}
+{-# LANGUAGE DeriveAnyClass     #-}
+{-# LANGUAGE DeriveGeneric      #-}
+{-# LANGUAGE PolyKinds          #-}
+{-# LANGUAGE StandaloneDeriving #-}
+{-# LANGUAGE FlexibleInstances  #-}
+
+------------------------------------------------------------------------
+
+-- Solution to the famous Tower of Hanoi puzzle using tools for state
```
stevana: Please add a module header like in the other examples:
```haskell
-----------------------------------------------------------------------------
-- |
-- Module      :  Hanoi
-- Copyright   :  (C) 2019, Adam Boniecki
-- License     :  BSD-style (see the file LICENSE)
--
-- Maintainer  :  Adam Boniecki <email@>
-- Stability   :  provisional
-- Portability :  non-portable (GHC extensions)
```

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+-- be placed on top of a smaller disc. Only one disc can be moved at a time.
+
+------------------------------------------------------------------------
+
+module Hanoi
+  ( Command(..)
+  , Model(..)
+  , initModel
+  , transitions
+  , prop_hanoi
```
stevana: Only export `prop_hanoi`, the rest are not used anywhere?

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+import Data.Maybe
+import Data.TreeDiff.Expr ()
+import GHC.Generics
+         (Generic, Generic1)
+import Prelude
+import Test.QuickCheck
+         (Gen, Property, (===), Arbitrary (arbitrary), choose, suchThat)
+import Test.QuickCheck.Monadic
+         (monadicIO)
+import Test.StateMachine
```
stevana: Newline here to separate imports of external packages from internal imports (this package, i.e. Test.StateMachine).

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+
+data Response (r :: Type -> Type) = Done
+  deriving (Show, Generic1, Rank2.Foldable)
+
+------------------------------------------------------------------------
+
+transitions :: Model r -> Command r -> Response r -> Model r
+transitions (Model pegs) (Move (from,to)) _ =
+  Model $ pegs // [(from, xs), (to, x:(pegs!to))]
+    where (x, xs) = (head pegs_from, tail pegs_from)
```
stevana: This can be simplified as follows:
```haskell
where
  (x :: xs) = pegs ! from
```

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+          pegs_from = pegs!from
+
+preconditions :: Model Symbolic -> Command Symbolic -> Logic
+preconditions (Model pegs) (Move (from,to)) = (Boolean $ isJust x) :&& (x .<= y)
+  where x = listToMaybe $ pegs!from
```
stevana: Please use space around operators such as `!`.

test/Spec.hs
```diff
@@ -44,6 +45,8 @@ tests :: Bool -> TestTree
 tests docker0 = testGroup "Tests"
   [ testCase "Doctest Z module" (doctest ["src/Test/StateMachine/Z.hs"])
   , ShrinkingProps.tests
+  , testProperty "Towers of Hanoi"
+      (expectFailure (prop_hanoi 3))
```
stevana: Does it always find the solution in 100 tries? If not please use `withMaxSuccess` like below to increase the number of tries.

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+        -- Any disc can be placed on empty peg, so no disc counts as largest disc
+        y = listToMaybe $ (pegs!to) ++ [maxBound]
```
stevana: Are the parenthesis needed here?

test/Hanoi.hs
```diff
@@ -0,0 +1,127 @@
...
+
+-- Check if all discs are at the last peg. The invariant states that this is not
+-- the case, so when it is not satisfied, we have a counter example that is a
+-- solution to our puzzle.
+
+postconditions :: Model Concrete -> Command Concrete -> Response Concrete -> Logic
+postconditions m c r = length lst ./= (sum $ fmap length pegs)
```
stevana: `./= sum (fmap length pegs)` is more readable, I think.

test/Hanoi.hs
abonie: I am not happy with my version either, but `(x : xs) = ...` caused a warning about non-exhaustive pattern match - I think it was safe, because precondition guarantees `pegs ! from` won't be `[]`, but still ugly (and not sure if shrinking maintains preconditions anyway?).
I will go with version below if that's ok
```haskell
transitions m@(Model pegs) (Move (from,to)) _ = case pegs ! from of
  (x : xs) -> Model $ pegs // [(from, xs), (to, x:(pegs ! to))]
  _ -> m
```

test/Spec.hs
abonie: With 3 discs yes, so far it have not failed to find one for me, seems reliable. For 4 discs it would probably need around 2000 tries.

test/Hanoi.hs
stevana: Shrinking preserves pre-conditions. Rather do `[] -> error "transition: impossible, due to preconditon" and just returning `m`.

Issue #15, opened 2017-03-16
Generalise to N threads in the parallel property


stevana:
Currently only two parallel workers are used in the parallel property, some bugs might require more things to happen in parallel...

Danten:
One idea I have for this is to replace `Fork` with the following:
```
data Fork t a = Fork {
  prefix :: a
  suffix :: t a
}
```
Where we would often require `t` to be `Traversable`. To get back the current behaviour we would use:
```
data Pair a = Pair a a

instance Traversable Pair where
  traverse f (Pair x y) = Pair <$> f x <*> f y

type OldFork a = Fork Pair a
```
In general we could have `Fork (Vec N)` to have `N` threads.

stevana:
I noticed that Erlang uses a list of suffices, i.e. `ParallelProg = { prefix: Prog, suffixes : [Prog] }`...

stevana:
I remember John Hughes quoting a paper that shows that empirically most concurrency bugs can be triggered with two threads.

It can become more important to have more threads when doing fault injection though. For example, Jepsen uses 5 threads.

kderme:
@stevana, as another generalisation, I think the type  of these threads could be extended. For example forking unbound vs bound threads could make a difference at the result, when testing some frameworks. It would be nice if the user had this option. 
Thoughts?

stevana:
Nice idea!

(Having a user-land scheduler that we could deterministically control would be great, but that's much more work.)

stevana:
I've found the quote again:

> "Almost all (96%) of the examined concurrency
bugs are guaranteed to manifest if certain partial order between 2
threads is enforced."

From the end of the "Parallelism in test cases" section in the paper "Finding Race Conditions in Erlang with QuickCheck and PULSE" which is linked to from the README. This section also talks about performance issues with multiple threads.

PR #294, opened 2019-04-07
Generalize to N threads

kderme:
This pr is WIP. Related issue https://github.com/advancedtelematic/quickcheck-state-machine/issues/15
So far Parallel Commands were executed by 2 threads. In this Pr I generalize this: each suffix, instead of a Pair of `Commands` is now a list of `Commands` and each thread executes one of them.

Things TODO/future improvements:
- [x] there is a bug at shrinking, which I have to fix before merging this pr. (more below).
- [x] test more examples. So far I have only tested MemoryReference.
- [x] test the shrinker by extending shrinking properties. We now test that the execution for n=1 is equivalent to the sequential and that shrinking for n=2 is equivalent to the existing parallel.
- [ ] optimize. Even for 4 threads, execution gets very slow.  Tuning parameters like suffix length may help. Edit: some thoughts I'm working on https://github.com/advancedtelematic/quickcheck-state-machine/issues/321
- [x] generalize also the ASCII diagram of parallel commands (see `toBoxDrawings`). Edit: We decided to use dot. Maybe on some different pr? Issue created: https://github.com/advancedtelematic/quickcheck-state-machine/issues/316
- [x] ~probably there are more ways to shrink.~ Edit: that's not needed since it will increase execution time and shrinking will not anymore be equivalent for n=2.
- [x] add comments and beautyfy.
- [ ] The pr is built so that it is NOT a breaking api change. This means however that currently there are 2 versions of each function: the existing one which works on Pairs and a new one which works on Lists. As discussed in the Issue 15, I could generalize the existing functions to work on some Traversable. I believe this will be an easy generalizion once the list works properly for all tests and the bug is fixed. Edit: Some internal functions already work on Foldable or Traversable.

About the BUG (Edit: Fixed):

The bug is on shrinking, probably at `shrinkAndValidateGeneralParallel`, somewhere in the combination of environments (I have left a comment in code). The bug some times appears, when I run prop_general_parallel of MemoryReference, with 2,3 or 4 threads and a Race Bug:
```
OK (0.66s)
      +++ OK, failed as expected. (after 9 tests and 3 shrinks):
      Exception:
        executeCommands: impossible 
        Left (EnvironmentValueNotFound (Var 0))
        CallStack (from HasCallStack):
          error, called at src/Test/StateMachine/Sequential.hs:336:33 in quickcheck-state-machine-0.6.0-34jzR3JpmPESvqdQjrvbB:Test.StateMachine.Sequential
      ParallelCommands
        { prefix =
            Commands
              { unCommands =
                  [ Command Create (Created (Reference (Symbolic (Var 0)))) [ Var 0 ]
                  ]
              }
        , suffixes =
            [ []
            , [ Commands
                  { unCommands =
                      [ Command (Read (Reference (Symbolic (Var 0)))) (ReadValue 0) [] ]
                  }
              ]
            ]
        }
      *** Exception running callback: 
      executeCommands: impossible 
      Left (EnvironmentValueNotFound (Var 0))
      CallStack (from HasCallStack):
        error, called at src/Test/StateMachine/Sequential.hs:336:33 in quickcheck-state-machine-0.6.0-34jzR3JpmPESvqdQjrvbB:Test.StateMachine.Sequential
```
The test seems to succeed (or maybe fail as expected), but it fails for the wrong reason. I am not quite sure why the example above throws the `EnvironmentValueNotFound`, I need to dig a bit deeper.

(By the way these kind of false positives makes me think `expectFailure` should fail in such cases, something you mention there https://github.com/nick8325/quickcheck/issues/230)

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -19,21 +19,27 @@
 -----------------------------------------------------------------------------
 
 module Test.StateMachine.Parallel
-  ( forAllParallelCommands
+  ( forAllNParallelCommands
+  , forAllParallelCommands
+  , generateNParallelCommands
   , generateParallelCommands
+  , shrinkNParallelCommands
   , shrinkParallelCommands
+  , shrinkAndValidateNParallel
   , shrinkAndValidateParallel
+  , shrinkCommands'
+  , runNParallelCommands
   , runParallelCommands
+  , runNParallelCommandsNTimes
   , runParallelCommandsNTimes
   , executeParallelCommands
   , linearise
   , toBoxDrawings
+  , prettyNParallelCommands
   , prettyParallelCommands
   , advanceModel
   ) where
 
-import           Control.Arrow
-                   ((***))
 import           Control.Monad
                    (foldM, replicateM)
 import           Control.Monad.Catch
@@ -43,8 +49,10 @@ import           Control.Monad.State.Strict
 import           Data.Bifunctor
                    (bimap)
 import           Data.List
-                   (partition, permutations)
+                   (find, partition, permutations)
 import qualified Data.Map.Strict               as Map
+import           Data.Maybe
+                   (fromMaybe)
 import           Data.Monoid
                    ((<>))
 import           Data.Set
@@ -62,7 +70,8 @@ import           Text.PrettyPrint.ANSI.Leijen
 import           Text.Show.Pretty
                    (ppShow)
 import           UnliftIO
-                   (MonadIO, MonadUnliftIO, concurrently, newTChanIO)
+                   (MonadIO, MonadUnliftIO, concurrently,
+                   forConcurrently, newTChanIO)
 
 import           Test.StateMachine.BoxDrawer
 import           Test.StateMachine.Logic
@@ -81,6 +90,18 @@ forAllParallelCommands :: Testable prop
                        -> Property
 forAllParallelCommands sm =
   forAllShrinkShow (generateParallelCommands sm) (shrinkParallelCommands sm) ppShow
+
+
+forAllNParallelCommands :: Testable prop
+                        => (Show (cmd Symbolic), Show (resp Symbolic), Show (model Symbolic))
+                        => (Rank2.Traversable cmd, Rank2.Foldable resp)
+                        => StateMachine model cmd m resp
+                        -> Int                                      -- ^ Number of threads
+                        -> (NParallelCommands cmd resp -> prop)     -- ^ Predicate.
+                        -> Property
+forAllNParallelCommands sm np =
+  forAllShrinkShow (generateNParallelCommands sm np) (shrinkNParallelCommands sm) ppShow
+
 
 -- | Generate parallel commands.
 --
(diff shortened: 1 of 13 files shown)


kderme:
Most of the times, it successfully shrinks
```
 ParallelCommands
        { prefix =
            Commands
              { unCommands =
                  [ Command Create (Created (Reference (Symbolic (Var 0)))) [ Var 0 ]
                  ]
              }
        , suffixes =
            [ [ Commands
                  { unCommands =
                      [ Command (Increment (Reference (Symbolic (Var 0)))) Incremented []
                      , Command (Read (Reference (Symbolic (Var 0)))) (ReadValue 1) []
                      ]
                  }
              , Commands
                  { unCommands =
                      [ Command (Write (Reference (Symbolic (Var 0))) 2) Written [] ]
                  }
              ]
            ]
        }
```

test/MemoryReference.hs
```diff
@@ -177,6 +178,12 @@ prop_parallel bug = forAllParallelCommands sm' $ \cmds -> monadicIO $ do
     where
       sm' = sm bug
 
+prop_general_parallel :: Bug -> Int -> Property
```
stevana:
```suggestion
prop_generalParallel :: Bug -> Int -> Property
```

The idiom to use the `prop_` snake case prefix comes from QuickCheck's `quickCheckAll` template Haskell code that scans the file for properties with that prefix, we should use Haskell's camel case for the rest of the name.

test/Spec.hs
```diff
@@ -46,11 +46,19 @@ tests docker0 = testGroup "Tests"
...
-      , testProperty "RaceBugSequential"                (prop_sequential Race)
-      , testProperty "RaceBugParallel"   (expectFailure (prop_parallel   Race))
-      , testProperty "PreconditionFailed" prop_precondition
-      , testProperty "ExistsCommands"     prop_existsCommands
+       , testProperty "LogicBug"          (expectFailure (prop_sequential Logic))
+       , testProperty "RaceBugSequential"                (prop_sequential Race)
+       , testProperty "RaceBugParallel"   (expectFailure (prop_parallel   Race))
+       , testProperty "PreconditionFailed" prop_precondition
+       , testProperty "ExistsCommands"     prop_existsCommands
+       , testProperty "NoBug 1 thread"           (prop_general_parallel None 1)
```
stevana: Perhaps you can check that `prop_general_parallel bug 1` is equivalent to `prop_sequential bug` and `prop_general_parallel bug 2` is equivalent `prop_parallel bug`? In the parallel case it might not be so easy because of the non-determinism of the scheduler, but perhaps you could check that the shrinking part at least matches up, similar to how it's done in: https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/test/ShrinkingProps.hs .

src/Test/StateMachine/Parallel.hs
```diff
@@ -299,6 +438,73 @@ shrinkAndValidateParallel sm = \shouldShrink env (ParallelCommands prefix suffix
...
+    case shouldShrink of
+      DontShrink -> concatMap (go' DontShrink) (shrinkAndValidate sm DontShrink env prefix)
+      MustShrink -> concatMap (go' DontShrink) (shrinkAndValidate sm MustShrink env prefix)
+                  ++ concatMap (go' MustShrink) (shrinkAndValidate sm DontShrink env prefix)
+  where
+    withCounterFrom :: ValidateEnv model -> ValidateEnv model -> ValidateEnv model
+    e `withCounterFrom` e' = e { veCounter = veCounter e' }
+
+    go :: Commands cmd resp          -- validated prefix
+        -> ValidateEnv model         -- environment after the prefix
```
stevana:
```suggestion
       -> ValidateEnv model         -- environment after the prefix
```

kderme:
review mostly addressed and added new examples. At the new examples the bug never appears, only on MemoryReference.

stevana:
> At the new examples the bug never appears, only on MemoryReference

The bug is probably related to references. Neither the echo nor ticket dispenser examples use references. The error encountered example does use them, but it also shortcuts generation quite early due to the `ErrorEncountered` state, which perhaps makes the bug manifest itself less often.

kderme:
@stevana I found some time to fix this bug. The issue was that while shrinking it's possible to shrink to less threads than initially specified, so you could end up with a suffix which is an empty list. In my generalisation, I use mconcat to combine the resulting environments of each thread, but when the list is empty, mconcat gives an empty environment, which later resulted in `EnvironmentValueNotFound` when the next suffix was executed,

test/Spec.hs
kderme: I don't think shrinking matches up atm. The difference is that when you shrink a pair of commands it always consists of 2 commands. However when you shrink a list of 2 it may actually result in a singleton or an empty list. I think I should change the shrinking so that it retains the length of the list and it rather only shrinks the elements. This will also give a better fix on the bug that appeared.

kderme: For example `shrinkPairS' shrinkListS' ([1,2],[3,4])` gives
```
[Shrunk {wasShrunk = True, shrunk = ([],[3,4])},Shrunk {wasShrunk = True, shrunk = ([2],[3,4])},
Shrunk {wasShrunk = True, shrunk = ([1],[3,4])},Shrunk {wasShrunk = True, shrunk = ([1,2],[])},
Shrunk {wasShrunk = True, shrunk = ([1,2],[4])},Shrunk {wasShrunk = True, shrunk = ([1,2][3])}, Shrunk {wasShrunk = False, shrunk = ([1,2],[3,4])}]
```
but `shrinkListS shrinkListS' [[1,2],[3,4]]` gives
```
[Shrunk {wasShrunk = True, shrunk = []},Shrunk {wasShrunk = True, shrunk = [[3,4]]},
Shrunk {wasShrunk = True, shrunk = [[1,2]]},
Shrunk {wasShrunk = True, shrunk = [[][3,4]]},Shrunk {wasShrunk = True, shrunk = [[2],[3,4]]},
Shrunk {wasShrunk = True, shrunk = [[1] [3,4]]},Shrunk {wasShrunk = True, shrunk = [[1,2],[]]},
Shrunk {wasShrunk = True, shrunk = [[1,2],[4]]},Shrunk {wasShrunk = True, shrunk = [[1,2],[3]]},
Shrunk {wasShrunk = False, shrunk = [[1,2],[3,4]]}]
```
You can see that the first three elements should be fitlered out, so that there is some sence of equivalence. Doing this filter makes shrinking less agressive, but maybe it is good as a first step to check the equivalence with the older implementation.

stevana: Hmm, sometimes too aggressive shrinking isn't good either. For example, if either of the two inner lists are empty then we are effectively executing the commands on a single thread.

stevana:
Regarding optimisation: there's a section called "Parallelism in test cases" in the paper [Finding Race Conditions in Erlang with QuickCheck and PULSE](http://www.cse.chalmers.se/~nicsma/papers/finding-race-conditions.pdf) that contains some useful analysis.

stevana:
Their setting is slightly different than ours though, as we execute batches of commands in parallel, then synchronise...

stevana:
Regarding visualisation: perhaps it's easier to use/read graphviz (dot) instead of ASCII?

kderme:
> Regarding visualisation: perhaps it's easier to use/read graphviz (dot) instead of ASCII?
> 
> Examples:
> 
> * Jepsen: https://github.com/jepsen-io/knossos/blob/master/README.md#visualization-of-faults
> * Molly: https://gist.github.com/cstorey/38d2ba04f661a7ed727d76cd47d6b40e#file-lamport-svg

Agree ASCII gets messed up sometimes.

kderme:
First commit is the one from here https://github.com/advancedtelematic/quickcheck-state-machine/pull/315 (Ci fails, but I think this is because some test takes too long..). Also shrinking properties are now added.

src/Test/StateMachine/Types.hs
```diff
@@ -117,3 +125,14 @@ toPair :: (a, a) -> Pair a
 toPair (x, y) = Pair x y
 
 type ParallelCommands = ParallelCommandsF Pair
+
+type NParallelCommands = ParallelCommandsF []
```
stevana: Hmm, I wonder if using length indexed lists, i.e. `Vec :: Nat -> Type`, instead of lists here would make things easier? Would make it possible to write `toPairUnsafe` in a safe way, but could also make other things more complicated...

stevana:
I agree that visualising counterexamples for N-parallel programs can be a separate PR.

src/Test/StateMachine/Types.hs
kderme: Indeed, I think writing shrinkers may be complicated. Also, I only use the unsafe functions in tests.

kderme:
Closed in favor of https://github.com/advancedtelematic/quickcheck-state-machine/pull/324

PR #296, opened 2019-04-09
[WIP] Fix/crud example docker

abonie:
There is a problem with example in `test/CrudWebserverDb.hs` where it will sometimes try connecting to postgres before the database is ready. This PR prevents this by adding another health check and also makes sure that docker containers and volumes are properly cleaned up afterwards.

The new health check creates some noisy output that obscures test results (that are hard to read anyway tbh), so this PR is WIP, as I hope to address this issue as well.


stevana:
Nice, thanks for fixing!

If the flakiness is gone, then please also do enable those tests in CI again (basically revert https://github.com/advancedtelematic/quickcheck-state-machine/commit/b239622fb09d277f01a43b8cd18a391eaabc3bc7)!

abonie:
I don't experience flakiness on my machine anymore, however tests from `test/CrudWebserverDb.sh` can take nearly 6 mins sometimes (usually much less, but rarely less than 120s)

PR #304, opened 2019-05-16
expose constructors for Markov and Transition

rdanitz:


stevana:
Lets not expose the internal structure of `Markov`, in case we want to change it later on.

I'd suggest we add a function: `markovToAdjacencyMap :: Markov state cmd_ prob -> Map state (Map state (cmd, prob))` instead. (This type is what the `alga` library uses for labelled graphs.)

src/Test/StateMachine/Markov.hs
```diff
@@ -20,7 +20,9 @@
 
 module Test.StateMachine.Markov
   ( Markov
+  , Transition
```
stevana: No need to export this anymore.

Issue #162, opened 2017-10-03
Towards Jepsen-like tests


stevana:
[Jepsen](https://github.com/jepsen-io/jepsen) is a framework for testing distributed systems. It has been used to find
bugs in [many](https://jepsen.io/analyses) systems, e.g. Kafka, Cassandra, RabbitMQ.

It is also based on performing randomly generate actions from multiple threads
and then using linearisation to ensure correctness. However it is different in
that it has a thread, called the Nemesis, which is solely dedicated to fault
injection. Fault injections include skewed clocks, gc/io pauses (killall -s
STOP/CONT), lost packets, timeouts, and network partitions between the
distributed servers.

In order to account for the faults, Jepsen has a richer notion of history than
ours which includes the possibily of operations failing or timing out. When an
action failes, we know for sure that it did not change the state of the system,
where as if an operation timed out we don't know for sure (the state could have
changed, but the server ack that didn't reach us).

It would be neat if we could perform similar tests. Exactly how this is supposed
to work is still not clear. Some answers can perhaps be found in the original
linearizabiliy paper or in the many blog posts and talks by "Aphyr", the author
of Jepsen, some of which are linked to in the README.

Perhaps a good start would be to only focus on failed actions to begin with.
This issue seems to pop up in some of our examples already, see discussion in
#159, and does not require a nemesis thread.

A first step might be to change the type of `Semantics` to account for possible
failures:

```haskell
data Result resp err = Ok resp | Fail err

type Semantics act m err = forall resp. act Concrete resp -> m (Result resp err)
```

(This can later be extended to deal with timeouts.)

The `ResponseEvent` constructor will need to be changed accordingly, and
`linearise` as well. I guess the right thing to do in linearise is to not update
the model and not check the post-condition. We could also change the
post-condition make assertions about `err` in addition to `resp`, but maybe this
is a refinement we can make when we see the need for it.

Thoughts?

stevana:
There's also the case when an action that returns a reference fails. In that case we probably want all actions that use the reference to fail as well. We can tell that an action uses a reference that failed to be created by the fact that isn't in the environment.

stevana:
I added the possibility for the semantics to fail in the branch `feat/failing-semantics`, haven't tested it yet (need to fix bracket first).

stevana:
@Danten pointed out that we probably want to flip the args to `Result` so that we can make it an instance of monad later.

stevana:
>  Crashes
>
[...]

PR #317, opened 2019-06-11
Towards being able to handle crashing processes.

stevana:
See #162 for details.

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -24,7 +24,9 @@ module Test.StateMachine.Parallel
   , shrinkParallelCommands
   , shrinkAndValidateParallel
   , runParallelCommands
+  , runParallelCommands'
   , runParallelCommandsNTimes
+  , runParallelCommandsNTimes'
   , executeParallelCommands
   , linearise
   , toBoxDrawings
@@ -35,7 +37,7 @@ module Test.StateMachine.Parallel
 import           Control.Arrow
                    ((***))
 import           Control.Monad
-                   (foldM, replicateM)
+                   (replicateM)
 import           Control.Monad.Catch
                    (MonadCatch)
 import           Control.Monad.State.Strict
@@ -54,7 +56,8 @@ import           Data.Tree
                    (Tree(Node))
 import           Prelude
 import           Test.QuickCheck
-                   (Gen, Property, Testable, choose, property, sized, forAllShrinkShow)
+                   (Gen, Property, Testable, choose, forAllShrinkShow,
+                   property, sized)
 import           Test.QuickCheck.Monadic
                    (PropertyM, run)
 import           Text.PrettyPrint.ANSI.Leijen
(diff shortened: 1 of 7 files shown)


kderme:
@stevana does this resolve the error `simplify: impossible, because of the structure of linearise.` which sometimes appear when there is some Exception?

stevana:
@kderme: no, that error should never happen -- please open a ticket if you see it.

src/Test/StateMachine/Parallel.hs
```diff
@@ -315,48 +327,99 @@ runParallelCommandsNTimes :: (Show (cmd Concrete), Show (resp Concrete))
...
+        -- It would be better if we made executeCommands take a Maybe Environment
+        -- instead of the Check...
+
+        (runStateT (executeCommands sm hchan (Pid 1) CheckNothing cmds1) (env, initModel, newCounter, initModel))
+        (runStateT (executeCommands sm hchan (Pid 2) CheckNothing cmds2) (env, initModel, newCounter, initModel))
+      go hchan ( reason1
+               , reason2
+               , env1 <> env2
+               ) pairs
+    go hchan (Ok, ExceptionThrown, env) (Pair cmds1 _cmds2 : pairs) = do
```
kderme: @stevana is there any reason to continue execution if there is an ExceptionThrown?

kderme: Sequential execution also stops if there is an exception.

stevana: I think we should stop execution for a specific process (or `Pid`) if there's an `ExceptionThrown` on that process (but continue executing on the other processes). I believe that's what I've implemented here.

This is also why we need your genernalisation to allow execution on more than two processes, so that we can introduce faults (exceptions) without stopping all execution altogether.

kderme: I see, I guess it depends on how we want to simulate user usage of the api and since this gives more user option I think it's on the right direction. @stevana do you think this is close to being merged? I see some errors like `executeCommands: impossible` and some false positives tests, which I believe will be solved with this pr.

stevana: Another important thing that this PR adds is the ability to _complete_ a history which contains crashed processes. We do so by appending a user specified response to the end of the history. For example in the case of the memory reference example, if a `Write` fails then we can have a history like this:

```
Process 1: |-- x <- Create --|                           |---- Read x ---|
Process 2:                       |---- Write 5 x  --...        
```
In this case `Read` can return both `0` (the default value of create) or `5` depending on if `Write` crashed before or after writing.

To account of this we  complete the above history as follows:
```
Process 1: |-- x <- Create --|                           |---- Read x ---|
Process 2:                       |---- Write 5 x  -------------------------|    
```

Notice that the `Write` is concurrent with the `Read`. Now because of the way `linearise` works (it tries all possible sequential interleavings of the action calls) it will accept `Read` returning both `0` and `5`.

To complete for `Write` is easy as it's just an `Ack`, and it doesn't really matter what we complete `Read` with as it doesn't change the model. If a `Create` crashes, we are kind of screwed that's why in this PR I also allowed the pre-condition failed exception to be thrown. No doubt will there be examples for which complete will be harder...

A possible alternative would be to have the user catch all exceptions (due to fault injection) and account of the non-determinism, e.g. have not just a single value for each memory reference, but a set of values. This complicates the model and is a lot more work for the user though.

Yet another possibility might be to have the fault injection be more precise, so that we know if the `Write` failed before or after it wrote to the memory. That way we know if it should return `0` or `5`. I'm not sure if having this precision is always possible. It also complicates the model because we need to keep track of exactly what faults are injected and return the correct response in the presence of the fault.

Does this make sense? If not, let me know where I lost you and I can try to explain further.

It's my current understanding of the Linearizability paper and what Jepsen does. It took me a while to get here, and I'm still not certain if I understand things correctly. So I think it would be great if: 1) I could convince you that this approach makes sense, and 2) that we develop more examples that confirm that it indeed works.

PR #319, opened 2019-06-13
Make simplify work for empty interleavings, fixes #318.

### LINKED ISSUE #318: simplify: impossible, because of the structure of linearise
It appeared when a project threw an exception

stevana:
Signed-off-by: Stevan Andjelkovic <stevana@users.noreply.github.com>


stevana:
@kderme: does this fix the problem?

kderme:
> @kderme: does this fix the problem?

Yes seems to fix it. Even if both threads get an exception it still works with something like `EitherC BotC BotC`

PR #324, opened 2019-06-20
Generalize parallelism and visualization

kderme:
This replaces https://github.com/advancedtelematic/quickcheck-state-machine/pull/294 which was getting old. I rebased on the last changes and added dot visualization.

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -39,6 +39,8 @@ module Test.StateMachine.Parallel
   , toBoxDrawings
   , prettyNParallelCommands
   , prettyParallelCommands
+  , prettyParallelCommandsWithOpts
+  , prettyNParallelCommandsWithOpts
   , advanceModel
   ) where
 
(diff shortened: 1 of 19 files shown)


src/Test/StateMachine/Parallel.hs
```diff
@@ -662,16 +663,20 @@ exists' xs p = exists xs p
 prettyParallelCommands :: (MonadIO m, Rank2.Foldable cmd)
                        => (Show (cmd Concrete), Show (resp Concrete))
                        => ParallelCommands cmd resp
+                       -> Maybe GraphOptions
                        -> [(History cmd resp, Logic)] -- ^ Output of 'runParallelCommands'.
                        -> PropertyM m ()
-prettyParallelCommands cmds =
-  mapM_ (\(hist, l) -> printCounterexample hist (logic l) `whenFailM` property (boolean l))
+prettyParallelCommands cmds mGraphOptions hist =
```
stevana: Could we rename this to `prettyParallelCommandWithOpts` and define a backwards compatible `prettyParalellCommands` in terms of the `WithOpts` one where we pass `Nothing` as options? That way we couldn't break the old API.

stevana:
Travis CI says `stack build --resolver=lts-11` fails when trying to build `graphviz`:

```
graphviz                > /tmp/stack5540/graphviz-2999.20.0.3/Data/GraphViz/Printing.hs:348:37: error:
graphviz                >     Ambiguous occurrence ‘<>’
graphviz                >     It could refer to either ‘PP.<>’,
graphviz                >                              imported from ‘Text.PrettyPrint.Leijen.Text.Monadic’ at Data/GraphViz/Printing.hs:(79,1)-(85,68)
graphviz                >                           or ‘Data.Semigroup.<>’,
graphviz                >                              imported from ‘Data.Semigroup’ at Data/GraphViz/Printing.hs:105:24-36
graphviz                >     |
graphviz                > 348 |   unqtDot (BScheme n l) = unqtDot n <> unqtDot l
```

Can we use an older version or avoid this problem some other way?

stevana:
Worst case we can also drop lts-11 support, but then we should remove it from .cabal (tested-with and change the base dependency constraint) and .travis.yaml.

kderme:
I found that graphviz-2999.18.0.1 works on lts-11.0, but not on 13.1. I'm searching if there is any way to conditionally add it.

stevana:
Another possibility might be to avoid using the `graphviz` package and emit the dot format directly as a string like we do [here](https://github.com/advancedtelematic/quickcheck-state-machine/blob/master/src/Test/StateMachine/Markov.hs#L374).

kderme:
I added a new stack.yaml for lts-11 and modified the travis command. Hope this will work.

PR #325, opened 2019-06-28
Add Overflow example which fails only with >=4 threads

kderme:

> Can you think of an example bug which only manifests itself when 3 or more threads are used? Would be > nice to have such an example, but doesn't have to be part of this PR.

This is an example which only manifests when there are >= 4 threads.

The Model is a set of references of Int8. Some commands pick a random reference and add a constant number (in an atomic way) and then substract the same number (in an atomic way). If there are many threads the result can overflow.

counter_example_4

Check also how commands are shrunk to sum to 128 (which is the minimum overflow point)

### DIFF
--- a/test/Overflow.hs
+++ b/test/Overflow.hs
@@ -1,0 +1,141 @@
+{-# LANGUAGE DeriveAnyClass       #-}
+{-# LANGUAGE DeriveGeneric        #-}
+{-# LANGUAGE ExplicitNamespaces   #-}
+{-# LANGUAGE FlexibleContexts     #-}
+{-# LANGUAGE FlexibleInstances    #-}
+{-# LANGUAGE MonoLocalBinds       #-}
+{-# LANGUAGE PolyKinds            #-}
+{-# LANGUAGE ScopedTypeVariables  #-}
+{-# LANGUAGE StandaloneDeriving   #-}
+{-# LANGUAGE UndecidableInstances #-}
+
+
+module Overflow
+  ( prop_sequential_overflow
+  , prop_parallel_overflow
+  , prop_nparallel_overflow
+  ) where
+
+import           Control.Concurrent (threadDelay)
+import           Control.Concurrent.MVar
+import           Control.Monad
+import           Data.Int
+import           GHC.Generics
+                   (Generic, Generic1)
+import           Prelude
+import           System.Random
+                   (randomRIO)
+import           Test.QuickCheck
+import           Test.QuickCheck.Monadic (monadicIO)
+import           Test.StateMachine
+import           Test.StateMachine.DotDrawing
+import           Test.StateMachine.Parallel
+import           Test.StateMachine.Types(Reference(..), Symbolic(..))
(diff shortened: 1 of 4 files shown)


test/Overflow.hs
```diff
@@ -0,0 +1,141 @@
...
+
+mock :: Model Symbolic -> Command Symbolic -> GenSym (Response Symbolic)
+mock _ cmd = case cmd of
+  Create           -> Created <$> genSym
+  Check _var       -> return $ IsNegative False
+  BackAndForth _ _ -> return $ Unit
+
+generator :: Model Symbolic -> Maybe (Gen (Command Symbolic))
+generator (Model []) = Just (return Create)
+generator m            = Just $ frequency
```
stevana: The `=` signs are not aligned.

test/Overflow.hs
```diff
@@ -0,0 +1,141 @@
...
+genCheck m = do
+  Check <$> elements (getVars m)
+
+genBackAndForth :: Model Symbolic -> Gen (Command Symbolic)
+genBackAndForth m = do
+  BackAndForth 50 <$> elements (getVars m)
+
+shrinker :: Model Symbolic -> Command Symbolic -> [Command Symbolic]
+shrinker _ (BackAndForth n var) = [ BackAndForth n' var | n' <- shrink n ]
+shrinker _ _             = []
```
stevana: `=` signs not aligned.

test/Overflow.hs
```diff
@@ -0,0 +1,141 @@
...
+           Nothing generator shrinker semantics mock
+
+prop_sequential_overflow :: Property
+prop_sequential_overflow = forAllCommands sm Nothing $ \cmds -> monadicIO $ do
+  (hist, _model, res) <- runCommands sm cmds
+  prettyCommands sm hist (res === Ok)
+
+prop_parallel_overflow :: Property
+prop_parallel_overflow = forAllParallelCommands sm $ \cmds -> monadicIO $ do
+    prettyParallelCommandsWithOpts cmds (Just $ GraphOptions "counter_example.png" Png) =<< runParallelCommands sm cmds
```
stevana: Very long line, perhaps put the options in a where block, i.e. where `opts = Just $ GraphOptions ...`?

test/Overflow.hs
```diff
@@ -0,0 +1,141 @@
...
+                , (3, genCheck m)
+                ]
+
+genCheck :: Model Symbolic -> Gen (Command Symbolic)
+genCheck m = do
+  Check <$> elements (getVars m)
+
+genBackAndForth :: Model Symbolic -> Gen (Command Symbolic)
+genBackAndForth m = do
+  BackAndForth 50 <$> elements (getVars m)
```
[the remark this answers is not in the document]
kderme: It could be within an interval, not completely arbitrary. Or else it won't fail only for 4 threads.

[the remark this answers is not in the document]
stevana: Ah yes! Yeah, an interval is fine and perhaps add a comment about it?

kderme:
Thanks for the review!

Issue #327, opened 2019-07-02
Fix CI


stevana:
* CrashAndLogicBugParallel sometimes fails, even with `maxSuccess 3000`...
* Builds timeout often (perhaps because a new release of stack invalidated the old cache)
* Generally more flaky since Travis got bought up

Possible solutions:

* Make a "fast test suite" that doesn't run all tests? Is the build too slow or does running the tests take too long?
* Disable tests temporarily, so that cache gets built, then enable tests again?
* Remove stack builds from CI?
* Investigate alternatives to Travis?

Thoughts?

kderme:
I thinks it's some tests which take too long. I can try a couple things to make them faster.

stevana:
[Here's](https://api.travis-ci.org/v3/job/553339915/log.txt) another problem that I've not been able to reproduce locally:
```
 NoRaceBug:                                                    FAIL
      Exception: QuickCheck.elements used with empty list
      CallStack (from HasCallStack):
        error, called at ./Test/QuickCheck/Gen.hs:203:15 in QuickCheck-2.13.1-1rO57X9FmbTEL9U6cBekOs:Test.QuickCheck.Gen
```

(Could't find any seed or `--quickcheck-replay` to reproduce this in the logs, shouldn't that always be there?)

kderme:
Is this the correct link you have above? I couldn't find the failure..

stevana:
I'm pretty sure (I doubled checked when I posted), could it be that because I reran the job (because it timed out) it overwrote the logs...?

stevana:
Hmm, https://travis-ci.org/advancedtelematic/quickcheck-state-machine/jobs/553776901:

```
...
Setting up build cache
$ export CASHER_DIR=${TRAVIS_HOME}/.casher
0.16s$ Installing caching utilities
0.00s1.18sattempting to download cache archive
fetching master/cache-linux-trusty-811a73447d636739ee8e0adc0d376798ed2e3294169c24b7738557e7b2e4b580.tgz
could not download cache
0.00s1.07sadding /home/travis/.ghc to cache
creating directory /home/travis/.ghc
adding /home/travis/.cabal to cache
creating directory /home/travis/.cabal
adding /home/travis/.stack to cache
creating directory /home/travis/.stack
adding /home/travis/build/advancedtelematic/quickcheck-state-machine/.stack-work to cache
creating directory /home/travis/build/advancedtelematic/quickcheck-state-machine/.stack-work
[...]

PR #328, opened 2019-07-02
fix ci times

kderme:

Related issue https://github.com/advancedtelematic/quickcheck-state-machine/issues/327
I think the issue is that ShrinkProps sometimes create too many references, so I used an upper limit. I believe this does not reduce too much the power of testing.

### DIFF
--- a/test/ShrinkingProps.hs
+++ b/test/ShrinkingProps.hs
@@ -60,10 +60,9 @@ import           Test.QuickCheck.Monadic
                    (monadicIO)
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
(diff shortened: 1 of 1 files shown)


stevana:
That does indeed seem to have improved things, thanks!

kderme:
There is still the issue that stack builds the whole Haskell, so it sometimes reaches the 50 mins limit. Maybe an easy fix is to build only the library with stack and not the tests.

stevana:
A new version of stack was recently released which invalidated Travis' cache, maybe now that we have a successful build and a new cache things might be faster again? Stack and cabal builds used to take roughly as long in the past...

PR #330, opened 2019-07-03
fix generators

kderme:
Related issue https://github.com/advancedtelematic/quickcheck-state-machine/issues/327

Note how this exception, which appears very often is most of the times hidden because of laziness:
```
> undefined `Prelude.elem` []
False
```

So the generation creates the exception many times, but the precondition hides it, since it never evaluates the undefined thunk.

### DIFF
--- a/test/CrudWebserverDb.hs
+++ b/test/CrudWebserverDb.hs
@@ -309,7 +309,8 @@ postconditions _         _              _              = error "postconditions"
 -- * How to generate and shrink programs.
 
 generator :: Model Symbolic -> Maybe (Gen (Action Symbolic))
-generator (Model m) = Just $ frequency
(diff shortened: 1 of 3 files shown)


kderme:
Not sure what actually forced the exception in the ci, but I think this fixes the problem.

stevana:
Ah yes, I thought we had fixed this, but it seems we only fixed it for the `MemoryReference` example (and in the readme).

The `ErrorEncountered` and `ShrinkingProps` examples seems to have the same problem, but we haven't seem them fail in the same way?

kderme:
> The ErrorEncountered and ShrinkingProps examples seems to have the same problem, but we haven't seem them fail in the same way?

I can also try to fix these other generators

Issue #340, opened 2019-07-22
Add pretty print for CounterExample and improve test output


kderme:
tests print things like 
```
	AnnotateC "Read" (PredicateC (1 :/= 2))
	PostconditionFailed "AnnotateC \"Read\" (PredicateC (6 :/= 5))" /= Ok
```
which may not be very helpful, or just hard to understand for the user

Also drawing for `CrashAndLogic` is broken for some reason.

stevana:
Yes, this can be improved for sure!

I once had the idea that a nice output could be obtained by combining the initial `Logic` formula and the `Counterexample` that the formula evaluated to, e.g. something like:

```haskell
prettyPrintCounterexample :: Logic -> Counterexample -> String
prettyPrintCounterexample l@(p :&& q) (Fst ce) = concat
  [ "When checking that: "
  , show l
  , " is true, "
  , "the first component is false, because: "
  , prettyPrintCounterexample p ce
  ]

eval :: Logic -> Either String ()
eval l = case logic l of
  VTrue -> Right ()
  VFalse ce -> Left (prettyPrintCounterexample l ce)
```

I think the tricky part is how to not be too verbose and yet precise. Perhaps underlining the component of the formula that was false? Perhaps one should check how deep the formula is and only show parts of the tree? What do other tools do?

PR #341, opened 2019-07-24
Show which exception was thrown

kderme:
Related issue: https://github.com/advancedtelematic/quickcheck-state-machine/issues/340
I have found this quite useful while debugging. It prints exactly which exception was thrown while running the semantics:
`ExceptionThrown "\"user error (failed)\"" /= Ok`
instead of
`ExceptionThrown /= Ok `

### DIFF
--- a/src/Test/StateMachine/Sequential.hs
+++ b/src/Test/StateMachine/Sequential.hs
@@ -377,7 +377,7 @@ executeCommands StateMachine {..} hchan pid check =
           case ecresp of
             Left err    -> do
               atomically (writeTChan hchan (pid, Exception err))
-              return ExceptionThrown
+              return $ ExceptionThrown err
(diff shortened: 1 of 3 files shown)


src/Test/StateMachine/Sequential.hs
```diff
@@ -377,7 +377,7 @@ executeCommands StateMachine {..} hchan pid check =
           case ecresp of
             Left err    -> do
               atomically (writeTChan hchan (pid, Exception err))
-              return ExceptionThrown
+              return $ ExceptionThrown $ show err
```
stevana: Is `show` needed here?

kderme: Yes sorry, not needed. Fixed.

PR #348, opened 2019-08-18
fix Mock in parallel

kderme:
This pr fixes what I think is an issue related to the mock function in the parallel case. It also adds a new example, which fails without the fix. The test is pretty simple: we model a system from which you can only try to create some references:
```
data Command r
  = Create
```
The system manages to create only the first reference:
```
data Response r
  = Created (Reference Int r)
  | NotCreated
```
The sequential case works well, but in the paralle case if two commands try to create a reference in parallel, like in the example below, we don't know which one will succeed.
```
      ParallelCommands
        { prefix = Commands { unCommands = [] }
        , suffixes =
            [ Pair
                { proj1 =
                    Commands
                      { unCommands =
                          [ Command Create (Created (Reference (Symbolic (Var 0)))) [ Var 0 ]
                          ]
                      }
                , proj2 =
                    Commands { unCommands = [ Command Create NotCreated [] ] }
                }
            ]
        }
```
This can result in a mismatch between the references of the Symbolic Responses and the concrete, which results in `MockSemanticsMismatch` error, because the execution cannot update the `Environment`. Not sure what the best fix is: In my fix I added an additional check in `parallelSafe`, which makes sure that in all permutations, the mocked responses return the same number of Vars.

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -182,7 +182,8 @@ generateParallelCommands sm@StateMachine { initModel } = do
 -- list of commands for which the preconditions of all commands hold
 -- for permutation of the list, i.e. it is parallel safe. The other
 -- half is the remainder of the input list.
-spanSafe :: StateMachine model cmd m resp
+spanSafe :: Rank2.Foldable resp
+         => StateMachine model cmd m resp
          -> model Symbolic -> [Command cmd resp] -> [Command cmd resp]
          -> ([Command cmd resp], [Command cmd resp])
 spanSafe _ _     safe []           = (reverse safe, [])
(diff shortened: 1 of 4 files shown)


kderme:
Again lts11 fails to find the cache and builds everything (build takes more than 40 mins). lts12 also fails to find the cache, but it makes it on time.

src/Test/StateMachine/Parallel.hs
```diff
@@ -238,18 +239,20 @@ generateNParallelCommands sm@StateMachine { initModel } np =
...
   . unCommands
   where
-    preconditionsHold _     []                              = True
-    preconditionsHold model (Command cmd resp _vars : cmds) =
+    preconditionsHold _     []                             = True
+    preconditionsHold model (Command cmd resp vars : cmds) =
         boolean (precondition model cmd) &&
-          preconditionsHold (transition model cmd resp) cmds
+          preconditionsHold (transition model cmd resp) cmds &&
+          length vars == length (getUsedVars $ fst $ runGenSym (mock model cmd) newCounter)
```
stevana: Could you please add a comment here about why this is needed and maybe also link to this PR?

kderme: > Could you please add a comment here about why this is needed and maybe also link to this PR?

Added comment and rebased to new master

PR #349, opened 2019-08-20
Write thread sqlite

kderme:
Add a new SQLite example. The semantics use SQLite in a bit different way than normal: all write request are serialized on a TBQueue and there is a single thread which takes them out of the queue and executes them.

I should note this test has some features borrowed from tests here https://github.com/input-output-hk/ouroboros-network and http://www.well-typed.com/blog/2019/01/qsm-in-depth/

### DIFF
--- a/test/Spec.hs
+++ b/test/Spec.hs
@@ -31,6 +31,7 @@ import           MemoryReference
 import           Overflow
 import           ProcessRegistry
 import qualified ShrinkingProps
+import           SQLite
 import           Test.StateMachine.Markov
                    (PropertyName, StatsDb, fileStatsDb)
 import           TicketDispenser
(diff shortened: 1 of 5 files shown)


kderme:
Failure is probably related to `runMigrationSilent` from persistent, which does some strange stuff with the stderr handle.

test/Spec.hs
```diff
@@ -119,6 +120,9 @@ tests docker0 = testGroup "Tests"
       , testProperty "3-threadsEquivFailingNoOp"
         $ expectFailure $ withMaxSuccess 1000 $ prop_nparallel_clean  3 (Eq True)  Cleanup.NoBug     NoOp
       ]
+  , testGroup "SQLite"
+      [ testProperty "Parallel" prop_parallel_sqlite
```
stevana: The sequential property is missing here.

kderme: That's intentianal. When I ran it localy, leaving one property fixed the issue with runMigrationSilent. I guessed this happened because Tasty runs properties in parallel. But in ci it failed even with one property :confused:

stevana: What happens if you use `runMigration` rather than `runMigrationSilent` (that seems to work in the CRUD example)?

kderme: It floods the test output, since I run the Migration on every test.

kderme: I changed to runMigration and fixed some other warnings.

PR #350, opened 2019-08-27
Rqlite test

kderme:
A test for a distributed db called [rqlite](https://github.com/rqlite/rqlite). It uses this [client](https://github.com/kderme/hs-rqlite) and uses docker. It also supports various injected errors, like killing/responding nodes, disconnecting/reconnecting nodes, pausing/unpausing, delays. Clock skews is wip.

### DIFF
--- a/src/Test/StateMachine/Parallel.hs
+++ b/src/Test/StateMachine/Parallel.hs
@@ -98,10 +98,11 @@ forAllParallelCommands :: Testable prop
                        => (Show (cmd Symbolic), Show (resp Symbolic), Show (model Symbolic))
                        => (Rank2.Traversable cmd, Rank2.Foldable resp)
                        => StateMachine model cmd m resp
+                       -> Maybe Int
                        -> (ParallelCommands cmd resp -> prop)     -- ^ Predicate.
                        -> Property
-forAllParallelCommands sm =
-  forAllShrinkShow (generateParallelCommands sm) (shrinkParallelCommands sm) ppShow
+forAllParallelCommands sm mminSize =
+  forAllShrinkShow (generateParallelCommands sm mminSize) (shrinkParallelCommands sm) ppShow
 
 
 forAllNParallelCommands :: Testable prop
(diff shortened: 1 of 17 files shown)


src/Test/StateMachine/Parallel.hs
```diff
@@ -159,9 +160,10 @@ generateParallelCommands :: forall model cmd m resp. Rank2.Foldable resp
                          => Show (model Symbolic)
                          => (Show (cmd Symbolic), Show (resp Symbolic))
                          => StateMachine model cmd m resp
+                         -> Maybe Int
                          -> Gen (ParallelCommands cmd resp)
-generateParallelCommands sm@StateMachine { initModel } = do
-  Commands cmds      <- generateCommands sm Nothing
+generateParallelCommands sm@StateMachine { initModel } mminSize  = do
```
stevana: Can we please document this as a breaking change in the changelog?

stack.yaml
```diff
@@ -10,6 +10,8 @@ extra-deps:
   - splitmix-0.0.2@sha256:e1c4d1202757d63cd4b4498d5fa600a792f2a315d2fdc6cf772f4679113c1819
   - tree-diff-0.0.2.1@sha256:4bd07bd26091f21710288c1037124bfe19aadb5bdf3df57c8374afaa5f14fef8
   - graphviz-2999.20.0.3
+  - git: https://github.com/kderme/hs-rqlite
```
stevana: The cabal based CI builds won't like this (adding it to cabal.project will only work with newer cabals). Could you make a new release perhaps?

test/RQlite.hs
```diff
@@ -0,0 +1,752 @@
...
+    c <- liftIO $ newMVar 0
+    ls <- runNParallelCommandsNTimes 1 (sm c $ Just Weak) cmds
+    prettyNParallelCommandsWithOpts cmds (Just $ GraphOptions "rqlite.png" Png) ls
+    liftIO $ print $ fst $ head ls
+    liftIO $ print $ interleavings $ unHistory $ fst $ head ls
+
+garbageCollect :: Model Concrete -> IO ()
+garbageCollect (Model _ nodes) =
+    forM_ nodes $ \(Reference (Concrete node), p) -> do
+        putStrLnM $ "stoping " ++ show p
```
stevana:
```suggestion
        putStrLnM $ "stopping " ++ show p
```

test/Spec.hs
```diff
@@ -124,6 +125,9 @@ tests docker0 = testGroup "Tests"
   , testGroup "SQLite"
       [ testProperty "Parallel" prop_parallel_sqlite
       ]
+  , testGroup "Rqlite"
+      [ testProperty "parallel" $ withMaxSuccess 10 $ prop_parallel_rqlite (Just Weak)
```
stevana: Could we please add a comment on why the sequential property isn't included here?

kderme:
Review addressed, thanks!
> Can we please document this as a breaking change in the changelog?

Probably there is some more work to be done in the changelog.

stevana:
Looks good now! Apart from lts-11, which I made a ticket about removing support for.

PR #355, opened 2019-09-08
High-level interface to the state machine API

edsko:
Submitting a draft PR to let you guys know we've been working on this and made quite a bit of progress. However, I am yet to use these abstractions in my real/large QSM test setup (or indeed in the QSM tests themselves that use the patterns from my blog post), so I'm sure a bit more further tweaking will be required. However, quite pleased with how this turned out.

This is joint work with Alfredo Di Napoli.

# High-level interface to the state machine API

 This captures the patterns described in http://www.well-typed.com/blog/2019/01/qsm-in-depth/ as a proper Haskell abstraction. The basic idea is that we have a concept of commands and responses (just like in the lower level API), along with two interpretors, one for the system under test and one for the model. We then compare that they can run in lockstep, comparing responses at every step, insisting that they must be equal.
    
Of course requiring that responses must be really _equal_ is too strong: the system under test might return a real file handle for example, where the model will return a mock file handle. So we compare responses up to a mapping from real handles to mock handles, which is maintained by the Lockstep infrastructure.
    
It comes in two variants: one n-ary variant supporting an arbitrary number of different types of handles, and one where there is precisely one. The n-ary version is needed if testing for example a system that might return both DB handles and file handles, but in most cases the simple one suffices and is much easier to use. The simple API is defined in terms of the n-ary one (which requires some more sophisticated types).
    
The `IORefs` test is a nice example that demonstrates quite how succinctly we can write model tests with these abstractions.

### DIFF
--- a/quickcheck-state-machine.cabal
+++ b/quickcheck-state-machine.cabal
@@ -34,6 +34,9 @@ library
                      , Test.StateMachine.ConstructorName
                      , Test.StateMachine.DotDrawing
                      , Test.StateMachine.Labelling
+                     , Test.StateMachine.Lockstep.Aux
+                     , Test.StateMachine.Lockstep.NAry
+                     , Test.StateMachine.Lockstep.Simple
                      , Test.StateMachine.Logic
                      , Test.StateMachine.Markov
                      , Test.StateMachine.Parallel
@@ -65,6 +68,7 @@ library
         process >=1.2.0.0,
         QuickCheck >=2.13.2,
         random >=1.1,
+        sop-core,
         split,
         text,
         tree-diff >=0.0.2.1,
@@ -133,6 +137,7 @@ test-suite quickcheck-state-machine-test
                        Echo,
                        ErrorEncountered,
                        Hanoi,
+                       IORefs,
                        MemoryReference,
                        Mock,
                        Overflow,
--- a/src/Test/StateMachine/Lockstep/Aux.hs
+++ b/src/Test/StateMachine/Lockstep/Aux.hs
@@ -1,0 +1,88 @@
+{-# LANGUAGE ConstraintKinds     #-}
+{-# LANGUAGE DataKinds           #-}
+{-# LANGUAGE FlexibleContexts    #-}
+{-# LANGUAGE GADTs               #-}
+{-# LANGUAGE KindSignatures      #-}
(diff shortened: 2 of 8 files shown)


edsko:
(This PR is really just one commit. The first commit is #354 .

stevana:
Cool, I like the two variants, simple and N-ary, and how short the example turned out!

edsko:
@stevana If you like the approach, if you prefer you could merge this as is. I think what's here works, and is correct, but it might not be quite general enough, we might need some additional hooks. (In particular, I want some hooks for labelling and for test setup.) I'm not sure when I'll get to that though. Up to you, whichever way you prefer.

stevana:
I'm happy to merge this, but it would be good to fix a couple of things that CI is complaining about:

  * generic-sop missing in stack lts-12: https://api.travis-ci.org/v3/job/582296479/log.txt
  * ghc 8.2.2 needs import Data.Monoid (<>): https://api.travis-ci.org/v3/job/582296475/log.txt

CI is a bit broken at the moment, I've created an issue for it -- but that's unrelated to this PR. Nevertheless we should probably fix CI before merging this... Do you need this released by the way?

edsko:
> Do you need this released by the way?

No, I probably can't quite use this in its currents state anyway, I'll need the generalizations. Merging would primarily be to avoid bit-rot I suppose.

edsko:
The 8.2 problem is easily fixed, but not entirely sure what to do about sop-core?

stevana:
Should also be an easy fix, the error message in the log explains it quite well:

```
In the dependencies for quickcheck-state-machine-0.6.0:

    sop-core needed, but the stack configuration has no specified version 

             (latest matching version is 0.5.0.0)

needed since quickcheck-state-machine is a build target.

Some different approaches to resolving this:

  * Recommended action: try adding the following to your extra-deps

    in /home/travis/build/advancedtelematic/quickcheck-state-machine/stack.yaml:

- sop-core-0.5.0.0@sha256:8734ab38b8c84837094eec657da0b58942e481e20166131f34cf6c7fe9787b07,2827
```

edsko:
Ah, okay, that works for you? You don't mind a dependency that's not in lts-12?

stevana:
Yeah, that's fine. (I think we already got other dependencies that are not part of Stackage.)

PR #357, opened 2019-11-17
Add link to Wire test, MGS course and C FFI circular buffer example.

stevana:


stevana:
Works, thanks @rdanitz!

stevana:
Another minimal state machine testing implementation: https://github.com/UlfNorell/statem-laws/blob/master/Registry.hs

Issue #356, opened 2019-09-09
Fix CI


stevana:
From https://travis-ci.org/advancedtelematic/quickcheck-state-machine/jobs/582287824 : 

```
AnnotateC "ExceptionThrown \"UnexpectedResponse \\\"peers were empty while querying status! This probably indicates that the node path /rqlite/file/data does not exist or the peers file was deleted\\\"\"" BotC

dot: runInteractiveProcess: runInteractiveProcess: exec: does not exist (No such file or directory)

FAIL (618.78s)

      *** Failed! Falsified (after 10 tests):

      ParallelCommands

        { prefix =

            Commands

              { unCommands =

                  [ Command

                      Spawn

                      0

                      1

                      s

                      1

                      s

                      Nothing

                      Resp

                      (Right (Spawned (Reference (Symbolic (Var 0)))))

                      [ Var 0 ]

                  , Command

                      Insert

                      (Reference (Symbolic (Var 0)))

                      Person { name = "Robert" , age = 88 }

                      Resp

                      (Right Unit)

                      []

                  ]

              }

        , suffixes =

            [ Pair

                { proj1 =

                    Commands

                      { unCommands =

                          [ Command

                              Spawn

                              1

                              1

                              s

                              1

                              s

                              (Just 0)

                              Resp

                              (Right (Spawned (Reference (Symbolic (Var 1)))))

                              [ Var 1 ]

                          ]

                      }

                , proj2 =

                    Commands

                      { unCommands =

                          [ Command

                              Insert

                              (Reference (Symbolic (Var 0)))

                              Person { name = "Kostas" , age = 45 }

                              Resp

                              (Right Unit)

                              []

                          , Command

                              Spawn

                              2

                              1

                              s

                              1

                              s

                              (Just 0)

                              Resp

                              (Right (Spawned (Reference (Symbolic (Var 2)))))

                              [ Var 2 ]

                          ]

                      }

                }

            , Pair

                { proj1 = Commands { unCommands = [] }

                , proj2 =

                    Commands

                      { unCommands =

                          [ Command Stop (Reference (Symbolic (Var 2))) Resp (Right Unit) []

                          ]

                      }

                }

            , Pair

                { proj1 =

                    Commands

                      { unCommands =

                          [ Command

                              ReSpawn

                              2

                              2

                              1

                              s

                              1

                              s

                              (Just 1)

                              Resp

                              (Right (Spawned (Reference (Symbolic (Var 3)))))

[...]

PR #358, opened 2020-01-02
Avoid unnecessary use of `run`

edsko:
The top-level of a QSM test typically looks something like this:

```haskell
forAllCommands ... $ \cmds -> QC.monadicIO $ do
  (hist, prop) <- .... runCommands ...
  prettyCommands ...
```

The problem is that `runCommands` lives in PropertyM. This makes it impossible to do any kind of bracketing. Fortunately, it turns out that this use of `PropertyM` is actually not needed at all. In this commit, I remove it without having to change much at all, except having to insert a few extra runs in the test. In particular, the above will now turn into

```haskell
forAllCommands ... $ \cmds -> QC.monadicIO $ do
  (hist, prop) <- .... run ( runCommands ... ) ...
  prettyCommands ...
```

which, crucially, would allow us to do something like

```haskell
forAllCommands ... $ \cmds -> QC.monadicIO $ do
  (hist, prop) <- .... run (bracket ... $ \r -> runCommands .. r .. ) ...
  prettyCommands ...
```

NOTE: I'm getting a test failure, but that exists on `master` already:

```
  Rqlite
    parallel:                                                      FAIL (0.30s)
      *** Failed! Falsified (after 1 test and 4 shrinks):
      ParallelCommands
        { prefix =
            Commands
              { unCommands =
                  [ Command
                      Spawn
                      0
                      1
                      s
                      1
                      s
                      Nothing
                      Resp
                      (Right (Spawned (Reference (Symbolic (Var 0)))))
                      [ Var 0 ]
                  ]
              }
        , suffixes = []
        }
      Use --quickcheck-replay=727617 to reproduce.
  ErrorEncountered
```

I don't think that's related to this PR at all though.

### DIFF
--- a/src/Test/StateMachine/Sequential.hs
+++ b/src/Test/StateMachine/Sequential.hs
@@ -35,6 +35,7 @@ module Test.StateMachine.Sequential
   , ShouldShrink(..)
   , initValidateEnv
   , runCommands
+  , runCommands'
   , getChanContents
   , Check(..)
   , executeCommands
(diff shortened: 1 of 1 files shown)


stevana:
Would it make sense to name the run function that doesn't use `PropertyM` as `runCommands' :: m ...` and then define the old run function `runCommands = run . runCommands' :: PropertyM m...`, and thus not break backwards compatibility?

> NOTE: I'm getting a test failure, but that exists on master already

Yeah, there's a ticket for this already #356

edsko:
> Would it make sense to name the run function that doesn't use PropertyM as runCommands' :: m ... and then define the old run function runCommands = run . runCommands' :: PropertyM m..., and thus not break backwards compatibility?

Yes, I suppose that would be sensible, if you are worried about breaking compatibility.

mrBliss:
> Would it make sense to name the run function that doesn't use `PropertyM` as `runCommands' :: m ...` and then define the old run function `runCommands = run . runCommands' :: PropertyM m...`, and thus not break backwards compatibility?

I have updated the PR to follow this approach.

stevana:
Sweet, lets get this merged!

mrBliss:
> Sweet, lets get this merged!

Yeah, let's merge this :+1:

PR #359, opened 2020-01-27
Relax QuickCheck bound

mrBliss:
In 81a6c1c8376d72c570bbfc65b626e67687c36d59 and a3050887cda78e216bc4222bccbc8594528cc637, the QuickCheck bound was bumped from 2.9.2 to 2.13.1 to 2.13.2, while 2.12 is already low enough.

### DIFF
--- a/quickcheck-state-machine.cabal
+++ b/quickcheck-state-machine.cabal
@@ -66,7 +66,7 @@ library
         time >=1.7,
         pretty-show >=1.6.16,
         process >=1.2.0.0,
-        QuickCheck >=2.13.2,
+        QuickCheck >=2.12,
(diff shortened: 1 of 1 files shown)


mrBliss:
I have compiled and tested with QuickCheck-2.12 and there are no regressions.

mrBliss:
> @rdanitz: Can you please merge the three open PRs? (I don't have admin rights any longer so I cannot override the failing CI.)

@rdanitz: Maybe you should give @stevana admin rights again?

rdanitz:
It was never supposed that @stevana was stripped of his rights in the first place ;) Now you should be able to merge again.

PR #360, opened 2020-01-27
CI Fixes

kderme:
Some fixes:
- .gitignore is impoved
- `persistent` version is bumped, to use the new `runMigrationQuiet`. The SQLite test now doesn't flow stdoutput.
- rqlite test only runs when docker is installed.
- The docker rqlite image version is specified. A newer version was recently released, but it breaks compatibility with `hs-rqlite` and makes the test fail consistently. ~Unfortunately, this doesn't fix the existing bug with rqlite test. So it's still possible to fail.~
- rqlite bug fixed

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ stack.yaml.lock
 cleanup-test-folder/
 sqlite-seq/
 sqlite-par/
+rqlite-test-output/
+*test-output.png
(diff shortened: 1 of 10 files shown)


stevana:
> A newer version was recently released, but it breaks compatibility with hs-rqlite and makes the test fail consistently. Unfortunately, this doesn't fix the existing bug with rqlite test. So it's still possible to fail.

So what's the plan here?

stevana:
It's green! Is this good to go, or still work in progress?

kderme:
Good to go for me :)

stevana:
> You’re not authorized to merge this pull request.

Geez, who put all this security in?! ;-(. Ping @rdanitz

PR #372, opened 2021-05-11
export toStateMachine from Test.StateMachine.Lockstep.NAry

duog:
The immediate motivation for this is to get access to `forAllCommands`,
to facilitate labeling of test cases.


edsko:
LGTM. This provides users with an escape hatch for functionality that isn't (yet) implemented in `Lockstep` -- though adding support for labelling to `Lockstep` itself would certainly be a valuable addition.

stevana (approved):

